PR #48, opened 2014-10-31
HtmlFormat.escape loop performance

jroper:
The following loop could probably be a bit faster if a `while` loop is used instead of a `foreach` expression. Might save some character boxing and the code would still be pretty simple.

```scala
  def escape(text: String): Html = {
    // Using our own algorithm here because commons lang escaping wasn't designed for protecting against XSS, and there
    // don't seem to be any other good generic escaping tools out there.
    val sb = new StringBuilder(text.length)
    text.foreach {
      case '<' => sb.append("&lt;")
      case '>' => sb.append("&gt;")
      case '"' => sb.append("&quot;")
      case '\'' => sb.append("&#x27;")
      case '&' => sb.append("&amp;")
      case c => sb += c
    }
    new Html(sb.toString)
  }
```

Cc @pvlugter.

### DIFF
--- a/api/src/main/scala/play/twirl/api/Formats.scala
+++ b/api/src/main/scala/play/twirl/api/Formats.scala
@@ -19,10 +19,47 @@ object Formats {
 
 /**
  * Content type used in default HTML templates.
+ *
+ * This has 3 states, either it's a tree of elements, or a leaf, if it's a leaf, it's either safe text, or unsafe text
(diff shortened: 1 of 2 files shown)


jroper:
Thanks @gklopper, that's good to know.

I think we may as well go ahead and merge this. @richdougherty what do you think?

richdougherty:
:+1:

PR #65, opened 2015-02-19
Add support for equivalence checks for BufferedContent objects

lloydmeta:
Came to realise that comparing Html objects does not work at the moment; not sure if this is a good way to do it or not, but just wanted to learn + discuss.

### DIFF
--- a/api/src/main/scala/play/twirl/api/Content.scala
+++ b/api/src/main/scala/play/twirl/api/Content.scala
@@ -57,4 +57,11 @@ abstract class BufferedContent[A <: BufferedContent[A]](protected val elements: 
 
   def body = builtBody
 
+  override def equals(obj: Any): Boolean = obj match {
+    case other: BufferedContent[_] if this.getClass == other.getClass => body == other.body
+    case _ => false
+  }
+
+  override def hashCode(): Int = this.getClass.hashCode() + body.hashCode()
+
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,3 +2,4 @@ language: scala
 script:
   - sbt +test +publishLocal plugin/test plugin/scripted
   - cd docs && sbt test validateDocs
+sudo: false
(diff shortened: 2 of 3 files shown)


jroper:
`hashCode` and `equals` should be implemented on the body, not on the elements, the fact that an `Html` object could be made up of many elements or not should be completely transparent to the outside world.  The following should be true, since they both correspond to exactly the same output:

```scala
Html(elements = Seq(Html("foo"), Html("bar"))) == Html(text = "foobar")
```

As currently implemented, they are not equal.

Also, this should be addressed at the `BufferedContent` level, so that it also works for xml and text templates.

lloydmeta:
> should be implemented on the body, not on the elements

> Also, this should be addressed at the BufferedContent level, so that it also works for xml and text templates.

Sounds good, thanks :) ; will re-do.

api/src/main/scala/play/twirl/api/Content.scala
```diff
@@ -57,4 +57,11 @@ abstract class BufferedContent[A <: BufferedContent[A]](protected val elements:
 
   def body = builtBody
 
+  override def equals(obj: Any): Boolean = obj match {
+    case other: BufferedContent[_] => body == other.body
```
jroper: Probably should make this also check that they are the same class, ie `other.getClass == this.getClass`, because `Xml("foo")` probably shouldn't equal `Html("foo")`.

A few tests would also be good.

lloydmeta: Right, added a check for the same class in the `equals` method and also in the `hashCode` method.

Tests also added for equality checking (let me know if you want tests for `hashCode` as well)

jroper:
Ok, this is great. One last thing, could you squash your commits?  For information on why/how, see [here](https://playframework.com/documentation/2.3.x/WorkingWithGit).

lloydmeta:
No problem. Squashed.

jroper:
Thanks @lloydmeta.

PR #67, opened 2015-03-12
Drop support for Scala 2.9

benmccann:


pvlugter:
This won't work while twirl is being cross-built against scala 2.9. The new method was added in 2.10. You can see this with:
```
sbt ++2.9.3 compile
```
```
[error] twirl/compiler/src/main/scala/play/twirl/compiler/TwirlCompiler.scala:458: value reportThrowable is not a member of scala.tools.nsc.interactive.Global
[error]           PresentationCompiler.global.reportThrowable(new Throwable("Timeout in getFunctionMapping"))
[error]                                       ^
[error] twirl/compiler/src/main/scala/play/twirl/compiler/TwirlCompiler.scala:463: value reportThrowable is not a member of scala.tools.nsc.interactive.Global
[error]               PresentationCompiler.global.reportThrowable(new Throwable("Throwable in getFunctionMapping", t))
[error]                                           ^
```

We could probably add this method implicitly just for scala 2.9, to get rid of the deprecation warning. But we may just drop scala 2.9 support anyway, as it doesn't support jdk 8.

jroper:
My vote is to drop Scala 2.9, the existing Twirl releases are more than adequate.

benmccann:
@jroper @pvlugter I updated this PR to start removing support for 2.9. The build seems to be failing for a reason that is unrelated to my change

pvlugter:
@benmccann try rebasing on master. A few more changes were needed for the bump to play 2.4.0-M3 in the docs project that you made the other day.

richdougherty:
I think some more changes are needed to `build.sbt` to remove the 2.9 cross-building.

I'm not sure about this, but should we bump to Twirl 1.1.0?

marcospereira:
@richdougherty sounds good to me.

marcospereira:
@pvlugter what do you think? Can we merge this PR?

jroper:
The PR as it stands only removes the scripted test. The build file needs to be updated to remove 2.9 cross building. Also the 2.9 specific source directories should be removed too.

marcospereira:
Closing this in favor of #74.

PR #76, opened 2015-05-08
Change template import priorities

richdougherty:
Imports are now added in different scopes so they can have different priorities. The lowest priority are the imports included with every template. Next in priority are the imports provided through the sbt plugin setting. Highest in priority are imports included in the template itself.

This fixes a regression in import behaviour from Play 2.3 to Play 2.4. However, there is one change. Play 2.3 had two levels of import priority, but this change introduces three levels, which should more often be what users want.

See https://github.com/playframework/playframework/issues/4366.


jroper:
Can you add a test that tests this:

```
@import foo.bar.Example
@(example: Example)

...
```

I think the imports probably had to be moved to where they are to support imports that are required by the parameter declaration, and that was the prime motivation for the change to the way imports worked.

richdougherty:
Added to the commit message:

> The code generation has changed slightly, with outer objects being generated if necessary and the template code being generated within a class in those objects. The static template object extends that class.
>
> For efficiency, the code generation takes account of whether or not there are  any extra imports and only generates extra scope objects if they are needed.

compiler/src/test/scala/play/twirl/compiler/test/CompilerSpec.scala
```diff
@@ -87,12 +87,36 @@ abstract class CompilerTests(oldParser: Boolean = false) extends Specification {
       set must be_==("firstsecondthird")
     }
 
-    "compile successfully (imports)" in {
```
richdougherty: @jroper: here's the test you were asking for. It was already there, I just hadn't run it. You're right that my changes had broken this use case. The new code should handle both this use case and also resolve the issue with ambiguous wildcard imports.

richdougherty: And here's the template that's being tested: https://github.com/richdougherty/twirl/blob/import-shadowing/compiler/src/test/resources/argImports.scala.html

PR #86, opened 2015-07-14
String interpolation support

julienrf:


julienrf:
I don’t understand what’s wrong with the build. Can anyone help me?

julienrf:
It seems that I can not document the API addition because the `docs` project does not depend on the `api` project.

jroper:
I think the problem is that travis is running two builds, one that builds the main project, the other that builds the docs project, the one that builds the docs project though isn't `publishLocal`ing the main project first, so it doesn't see the changes you've made.

jroper:
Fixed in #87 - once that is merged, we can rerun this build.

julienrf:
Thanks, now Travis is happy!

PR #88, opened 2015-09-05
Use unbescape for increased escaping performance

pauldraper:
Benchmarks for common escaping libraries at https://github.com/unbescape/unbescape-tests. (Some dependencies are a couple minor version behind. You can update them in pom.xml, though I saw no difference.)

Run them with

```sh
$ mvn3 exec:java -Dexec.mainClass=$BENCHMARK
```

where `$BENCHMARK` is html, xml, or javascript (or one of the other several formats that it supports). You can also see how each handles Unicode escapes, named character references, etc.

For me, unbescape was

**HTML**
* 2-14x the performance of commons-lang
* 3-8x the performance of esapi
* 0.5-1x the performance of spring-web

**XML**
* 3-15x the performance of commons-lang
* 2-12x the performance of esapi

**JS**
* 2-15x the performance of commons-lang

The higher differences are for inputs that needs no escaping. unbescape avoids `String` allocations entirely in that case.

The twirl HTML benchmarks came out to be the same as far as I could tell, though they don't test much.

In a production system, I saw 2x performance gain in templating large amounts of data in XML after switching to a custom `Format` using unbescape.

---

Have you considering using a benchmark framework, like ScalaMeter?

I wanted to add a couple benchmarks, but I wasn't sure what the best way was.

### DIFF
--- a/api/src/main/scala/play/twirl/api/Formats.scala
+++ b/api/src/main/scala/play/twirl/api/Formats.scala
@@ -3,7 +3,9 @@
  */
 package play.twirl.api
 
-import org.apache.commons.lang3.StringEscapeUtils
+import org.unbescape.html.HtmlEscape
+import org.unbescape.javascript.JavaScriptEscape
+import org.unbescape.xml.XmlEscape
 import scala.collection.immutable
 
 object MimeTypes {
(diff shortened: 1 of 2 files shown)


api/src/main/scala/play/twirl/api/Formats.scala
```diff
@@ -93,9 +58,7 @@ object HtmlFormat extends Format[Html] {
   /**
    * Creates a safe (escaped) HTML fragment.
    */
-  def escape(text: String): Html = {
-    new Html(Nil, text, true)
-  }
+  def escape(text: String): Html = Html(HtmlEscape.escapeHtml4(text))
```
pauldraper: Note: this *does* escape apostrophes.

The difference of HTML4 vs HTML5 is what named character references will be used (HTML5 having more).

pauldraper:
I'll need to revisit this. One test fails due to using a hexadecimal rather than a decimal escape.

The other fails due to escaping non-ASCII.

jroper:
The existing twirl html implementation delayed the escaping of the HTML until the full HTML was rendered - this means that if the string does need to be escaped, it saves two buffer copies.  The changes you've made, if there is some escaping that needs to be done, presumably do two unnecessary buffer copies.  So in the case where a lot of data that requires escaping is in the template, this is going to be a performance regression.

Happy to accept the changes for XML and JSON, but not for HTML, unless unbescape can escape directly into a `StringBuilder` to avoid the buffer copies.

schmitch:
I'm closing this since we now have a escape functionality which needs to work against scala-js, too.

PR #90, opened 2015-12-11
Update Reference from Play 2.3 to 2.4

arnzel:


wsargent:
Thanks!

mkurz:
Instead of updating the version in the URL you could just use "latest". E.g.
https://www.playframework.com/documentation/latest/ScalaTemplates

PR #98, opened 2016-03-25
Wrong error source position mapping result on Windows.

gslowikowski:
Position in source template file calculated from generated Scala file works good only if new line sequences contain only one char . On Windows it's two-char sequence `\r\n` and this is not handled properly.

When loading the generated file `TwirlIO.readFileAsString` method is used and then content is split into lines using `split('\n')` method. This works well. See `compiler/src/main/scala/play/twirl/compiler/TwirlCompiler.scala`:
- `GeneratedSource.content` method:
```scala
  def content = TwirlIO.readFileAsString(file, codec)
```
https://github.com/playframework/twirl/blob/1.1.1/compiler/src/main/scala/play/twirl/compiler/TwirlCompiler.scala#L111
and
- `GeneratedSource.toSourcePosition` method:
```scala
      val line = TwirlIO.readFileAsString(source.get, codec).substring(0, targetMarker).split('\n').size
```
https://github.com/playframework/twirl/blob/1.1.1/compiler/src/main/scala/play/twirl/compiler/TwirlCompiler.scala#L121

When converting source file offset to (line, column) pair different `IO.readLines` method is used to load file content into lines sequence. See `sbt-twirl/src/main/scala/play/twirl/sbt/TemplateProblem.scala`:
```scala
      val lines = source.toSeq flatMap { file => IO.readLines(file, codec.charSet) }
```
https://github.com/playframework/twirl/blob/1.1.1/sbt-twirl/src/main/scala/play/twirl/sbt/TemplateProblem.scala#L90

This method uses `BufferedReader` under the hood and does not work properly here. You don't know, how many new line characters are consumed during `readLine` method calls. `TwirlIO.readFileAsString(file, codec).split('\n')` should be used.

When reading lines with a Reader you don't know, if new line sequences contain one (Linux) or two (Windows) characters. Line/position calculated from offset using this method is wrong on Windows. `TwirlIO.readFileAsString(file, codec).split('\n')` has to be used instead.

Additionally `\r` character has to be removed from the end of the line (it was consumed by `IO.readLines`, but it isn't with the new method).

This bug was found and solution well tested when implementing Play! Framework source position mappers in my [Maven plugin](https://github.com/play2-maven-plugin/play2-maven-plugin/issues/92).

### DIFF
--- a/sbt-twirl/src/main/scala/play/twirl/sbt/TemplateProblem.scala
+++ b/sbt-twirl/src/main/scala/play/twirl/sbt/TemplateProblem.scala
@@ -5,6 +5,7 @@ package play.twirl.sbt
 
 import sbt._
 import play.twirl.compiler.{ GeneratedSource, MaybeGeneratedSource }
+import play.twirl.parser.TwirlIO
 import xsbti.{ CompileFailed, Maybe, Position, Problem, Severity }
 import scala.io.Codec
 
@@ -48,7 +49,7 @@ object TemplateProblem {
   class TemplatePosition(source: Option[File], location: Option[TemplateMapping.Location]) extends Position {
     val line: Maybe[Integer] = maybe { location map (_.line) }
 
-    val lineContent: String = location.fold("")(_.content)
+    val lineContent: String = location.fold("")(_.content.stripSuffix("\r"))
 
     val offset: Maybe[Integer] = maybe { location map (_.offset) }
 
@@ -87,7 +88,7 @@ object TemplateProblem {
     }
 
     def apply(source: Option[File], codec: Codec): TemplateMapping = {
-      val lines = source.toSeq flatMap { file => IO.readLines(file, codec.charSet) }
+      val lines = source.toSeq flatMap { file => TwirlIO.readFileAsString(file, codec).split('\n') }
       TemplateMapping(lines)
     }
   }


typesafehub-validator:
Hi @gslowikowski,

Thank you for your contribution! We really value the time you've taken to put this together.

Before we proceed with reviewing this pull request, please sign the Typesafe Contributors License Agreement:

[http://www.lightbend.com/contribute/cla](http://www.lightbend.com/contribute/cla)

sbt-twirl/src/main/scala/play/twirl/sbt/TemplateProblem.scala
```diff
@@ -87,7 +88,7 @@ object TemplateProblem {
     }
 
     def apply(source: Option[File], codec: Codec): TemplateMapping = {
-      val lines = source.toSeq flatMap { file => IO.readLines(file, codec.charSet) }
+      val lines = source.toSeq flatMap { file => TwirlIO.readFileAsString(file, codec).split('\n') }
```
marcospereira: Better to use `System.lineSeparator()` as the `split` parameter.

gslowikowski: Actually no. This will not work. You can use different new line separators. This depends on Git, SVN, etc. client configuration, IDE or other editor configuration.
I personally always try to use UNIX new lines when possible to avoid problems. Often have mixed after checkout with UNIX ones and some edits with an editor where it's not configurable.

gslowikowski:
After some mote tests it seems SBT removes trailing '\r' characters, so `stripSuffix("\r")` is redundant. I will investigate, what exactly SBT does.

gslowikowski:
... No, `stripSuffix("\r")` is not redundant. It's required here. SBT strips it somehow only when logging to console. It doesn't do it when redirecting output to a file.

sbt-twirl/src/main/scala/play/twirl/sbt/TemplateProblem.scala
```diff
@@ -48,7 +49,7 @@ object TemplateProblem {
   class TemplatePosition(source: Option[File], location: Option[TemplateMapping.Location]) extends Position {
     val line: Maybe[Integer] = maybe { location map (_.line) }
 
-    val lineContent: String = location.fold("")(_.content)
+    val lineContent: String = location.fold("")(_.content.stripSuffix("\r"))
```
gmethvin: I would just create `TemplateMapping#lines` like:
```scala
val lines: Seq[Line] = sourceLines.scanLeft(Line(0, -1, -1, "")) { (previous, content) =>
  Line(previous.line + 1, previous.end + 1, previous.end + 1 + content.length, content.stripSuffix("\r"))
}.drop(1)
```

sbt-twirl/src/main/scala/play/twirl/sbt/TemplateProblem.scala
gmethvin: `TemplateMapping` also assumes a new line is one character, so the position in the file won't line up with the line start anymore.

Actually I think it would be better for `TemplateMapping`'s constructor to accept the whole file string. The fact that it accepts lines and assumes one-character newlines seems like bad API. That change would introduce some binary compatibility issues so we'd probably want to do it separately, though.

gslowikowski: You can fix it differently. I tried to do it as simple as possible, without breaking any API.

BTW, there is `stripLineEnd` method in `StringLike` class, which looks a little bit nicer than `stripSuffix`. It's used in SBT [here](https://github.com/sbt/sbt/blob/v0.13.11/compile/interface/src/main/scala/xsbt/DelegatingReporter.scala#L61).

PR #100, opened 2016-04-08
Injectable templates

### LINKED ISSUE #71: Allow option to not import TemplateMagic._ in templates by default
I see here that the import is hard-coded: https://github.com/playframework/twirl/blob/e36a84922f933bb441b736560c7a6fb22f2b934a/compiler/src/main/scala/play/twirl/compiler/TwirlCompiler.scala#L332

In our case, this import is not desired and ends up conflicting with our own internal `RichDate` class. It would be nice to be able to configure Twirl to not include that import by default.

### LINKED ISSUE #93: Injectable templates
One problem with eliminating global state in Play is that now everything, eg CSRF, i18n, reverse routing, assets rendering etc needs to be passed into each template every time its invoked.  This is manageable (though not ideal) in Scala, but for Java users it leaves a lot to be desired.  One solution to this is to make templates themselves injectable.  I propose the following syntax:

```scala
@*
 * Some template
 *@
@this(templateContext: Context)

@(title: String)

...
```

The `this` keyword will be for defining constructors.  `this` is already a reserved word in Twirl by virtue of the fact that it's a Scala reserved word, and the `this()` syntax is somewhat consistent with Scala constructor syntax.  The above template would compile (roughly) to this:

```scala
class someTemplate @Inject() (templateContext: Context) extends Template1[String] {
  def apply(title: String) = { ... }
}
```

while, if no `@this` directive is specified, it will be compiled to an `object` as before.  Hence this should be 100% backwards compatible.

Templates could of course depend on each other, by declaring each other in their constructors, for example, the `index` template might depend on the `main` template:

**main.scala.html**:

```scala
@this(templateContext: Context)

@(title: String)(content: Html)

<html>
  <head><title>@title</title></head>
  <body>@content</body>
</html>
```

**index.scala.html**:

```scala
@this(templateContext: Context, main: views.html.main)

@main("Index") {
  <h1>Hello world!</h1>
}
```

Note that of course the `Context` type above is just a fictitious type that we may provide in Play as a convenience to provide web template utilities in one place, such as i18n, assets reverse routing/CDN/finger printing configuration, CSRF config, etc.

jroper:
Requires #99 to be merged first.

Fixes #93 Fixes #71.

Adds support for injectable templates.

The syntax for this is:

    @this(...)

Multiple parameter lists and implicit parameters are supported.  The last comment before the constructor is used as the comment for the constructor in the generated scala code.

In addition to implementing this, I greatly improved the way start of the file is passed - you can now put comments between import statements, and as many comments as you want between the top import statements and the parameter declaration (previously, you could only have one comment there).  

There is one breaking change that I had to make to do this - previously imports were scoped using a nested object pattern, allowing imports in the templates to override the imports defined in sbt, and those imports to override the default imports.  This feature has been removed, all these imports are in the same scope.  To compensate, I've done the following:

* There are now no default hard coded imports, all imports are configured in sbt. This means they can be completely overridden.  This alleviates the need to have the default imports in a lower priority scope to the sbt imports.
* I've greatly limited what gets imported, there is no longer a blanket import on play.twirl.api._.  Only things that are needed are imported.
* I've deprecated TemplateMagic (it's not magic, bad name), and it no longer gets imported.  It's kept around for binary compatibility (so templates compiled with the old Twirl will still work).
* I've replaced TemplateMagic with two objects, TwirlFeatureImports, for features of the Twirl language, and TwirlHelperImports, for helpers.  In addition, all the symbols exposed by importing these two have been namespaced, so there should be no naming collisions, making it less necessary to scope imports.
* In general, going forward, we should discourage using imports defined in sbt, and limit what Play adds, since it generally introduces more confusion than not (why do I need to import this, but not something in the models package?)  Using Twirl should be closer to regular Scala in its rules and imports, the only things automatically imported should be things that are generally considered features of the Twirl language.  If this is followed, scoping imports shouldn't be needed.

### DIFF
--- a/api/src/main/scala/play/twirl/api/TwirlFeatureImports.scala
+++ b/api/src/main/scala/play/twirl/api/TwirlFeatureImports.scala
@@ -1,0 +1,54 @@
+/*
+ * Copyright (C) 2009-2014 Typesafe Inc. <http://www.typesafe.com>
+ */
+package play.twirl.api
+
+import scala.language.implicitConversions
+
+/**
+  * Imports that provide Twirl language features.
+  *
+  * This includes:
+  *
+  * - @defining
+  * - @using
+  * - iterable/option/string as boolean for if statements
+  * - default values (maybeFoo ? defaultFoo)
+  */
+object TwirlFeatureImports {
+
+  /**
+    * Provides the `@defining` language feature, that lets you set a local val that can be reused.
+    *
+    * @param t The defined val.
+    * @param handler The block to handle it.
+    */
+  def defining[T](t: T)(handler: T => Any): Any = {
+    handler(t)
+  }
+
+  /** Provides the `@using` language feature. */
+  def using[T](t: T): T = t
+
+  /** Adds "truthiness" to iterables, making them false if they are empty. */
+  implicit def twirlIterableToBoolean(x: Iterable[_]): Boolean = x != null && !x.isEmpty
+  /** Adds "truthiness" to options, making them false if they are empty. */
+  implicit def twirlOptionToBoolean(x: Option[_]): Boolean = x != null && x.isDefined
+  /** Adds "truthiness" to strings, making them false if they are empty. */
+  implicit def twirlStringToBoolean(x: String): Boolean = x != null && !x.isEmpty
+
+  /**
+    * Provides default values, such that an empty sequence, string, option, false boolean, or null will render
+    * the default value.
+    */
+  implicit class TwirlDefaultValue(default: Any) {
+    def ?:(x: Any): Any = x match {
+      case "" => default
+      case Nil => default
+      case false => default
+      case 0 => default
+      case None => default
+      case _ => x
+    }
+  }
+}
--- a/api/src/main/scala/play/twirl/api/TemplateMagic.scala
+++ b/api/src/main/scala/play/twirl/api/TemplateMagic.scala
@@ -5,6 +5,7 @@ package play.twirl.api
 
 import scala.language.implicitConversions
 
+@deprecated("Use TwirlDefaultImports instead", "1.2.0")
 object TemplateMagic {
 
   // --- UTILS
--- a/api/src/main/scala/play/twirl/api/TwirlHelperImports.scala
+++ b/api/src/main/scala/play/twirl/api/TwirlHelperImports.scala
@@ -1,0 +1,40 @@
+/*
+ * Copyright (C) 2009-2014 Typesafe Inc. <http://www.typesafe.com>
+ */
+package play.twirl.api
+
+import scala.language.implicitConversions
+
+/**
+  * Imports for useful Twirl helpers.
+  */
+object TwirlHelperImports {
+
+  /** Allows Java collections to be used as if they were Scala collections. */
+  implicit def twirlJavaCollectionToScala[T](x: java.lang.Iterable[T]): Iterable[T] = {
+    import scala.collection.JavaConverters._
+    x.asScala
+  }
+
+  /** Allows inline formatting of java.util.Date  */
+  implicit class TwirlRichDate(date: java.util.Date) {
+
+    def format(pattern: String): String = {
+      new java.text.SimpleDateFormat(pattern).format(date)
+    }
+
+  }
+
+  /** Adds a when method to Strings to control when they are rendered. */
+  implicit class TwirlRichString(string: String) {
+
+    def when(predicate: => Boolean): String = {
+      predicate match {
+        case true => string
+        case false => ""
+      }
+    }
+
+  }
+
+}
--- a/compiler/src/main/scala/play/twirl/compiler/TwirlCompiler.scala
+++ b/compiler/src/main/scala/play/twirl/compiler/TwirlCompiler.scala
@@ -11,11 +11,12 @@ import play.twirl.parser.{TwirlIO, TwirlParser}
 
 object Hash {
 
-  def apply(bytes: Array[Byte], imports: String): String = {
+  def apply(bytes: Array[Byte], imports: Seq[String]): String = {
     import java.security.MessageDigest
     val digest = MessageDigest.getInstance("SHA-1")
     digest.reset()
-    digest.update(bytes ++ imports.getBytes)
+    digest.update(bytes)
+    imports.foreach(i => digest.update(i.getBytes("utf-8")))
     digest.digest().map(0xFF & _).map { "%02x".format(_) }.foldLeft("") { _ + _ }
   }
 
@@ -110,7 +111,7 @@ case class GeneratedSource(file: File, codec: Codec = TwirlIO.defaultCodec) exte
 
   def content = TwirlIO.readFileAsString(file, codec)
 
-  def needRecompilation(imports: String): Boolean = !file.exists ||
+  def needRecompilation(imports: Seq[String]): Boolean = !file.exists ||
     // A generated source already exist but
     source.isDefined && ((source.get.lastModified > file.lastModified) || // the source has been modified
       (meta("HASH") != Hash(TwirlIO.readFile(source.get), imports))) // or the hash don't match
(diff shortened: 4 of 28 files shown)


julienrf:
LGTM

api/src/main/scala/play/twirl/api/TwirlFeatureImports.scala
```diff
@@ -0,0 +1,46 @@
...
+  * Imports that provide Twirl language features.
+  *
+  * This includes:
+  *
+  * - @defining
+  * - @using
+  * - iterable/option/string as boolean for if statements
+  * - default values (maybeFoo ? defaultFoo)
+  */
+object TwirlFeatureImports {
```
gmethvin: Can we add type annotations on all these methods?

jroper: Added.

jroper:
@gmethvin when upgrading to this in Play, the `TwirlKeys.constructorAnnotations` should be added to the Play sbt plugin to include `@javax.inject.Inject()`.

gmethvin:
@jroper Thanks. I'm assuming this should be released as version 1.2.0? Or should we bump to 2.0.0?

mkurz:
I assume we can use injectable templates with Play 2.5 already? Is there something to watch out for?

johdah:
Is there any way to test this now? Can't resolve 1.2.0-SNAPSHOT

mkurz:
@johdah 
Add
```
resolvers += Resolver.sonatypeRepo("snapshots")
addSbtPlugin("com.typesafe.sbt" % "sbt-twirl" % "1.2.0-2016-06-04-cca9cd8-SNAPSHOT")
```
to your `plugins.sbt`.
You can find the snapshots here (scroll to the bottom to see the latest): https://oss.sonatype.org/content/repositories/snapshots/com/typesafe/sbt/sbt-twirl_2.10_0.13/

mkurz:
I tested the snapshot and until now it looks like everything works - also injected a couple of views.

johdah:
@mkurz Ah, I hade only added the resolver to build.sbt, not plugins.sbt. Thanks for the reminder.

johdah:
Are the views expected to be loaded in another way after this? Because I get many of the following error:

```
non-static method render(java.lang.String, java.lang.String) cannot be referenced from a static context
```

For example when I am using the view like this:

```java
return ok(views.html.about.render(user.email, user.company.name));
```

gmethvin:
@johdah the point of this change was to make the template into an injectable class rather than a static object. So now you need to inject an instance of `views.html.about` and then call `render` on the instance.

Maybe there should be an example of how to use it in the docs just to make it really clear.

johdah:
@gmethvin That's probably a good idea. So I inject an instance of the views now and am able to compile the project. But when I run it I get the following error:

```

**Comment by gmethvin** [comment]
@johdah See James's [comment](https://github.com/playframework/twirl/pull/100#issuecomment-217322197). You need to tell twirl to add the constructor annotation:
```scala
TwirlKeys.constructorAnnotations += "@javax.inject.Inject()"
```

When we add this to Play, we can add this setting for you automatically, since we know Play depends on javax.inject.

johdah:
@gmethvin I thought I was supposed to add that to build.sbt but that didn't work.

gmethvin:
@johdah "didn't work" in what way? You might want to check if the generated scala class (e.g. `target/scala-2.11/twirl/main/views/html/index.template.scala`) has the `@Inject` annotation on the constructor.

But yes, you need to add the constructor annotations to build.sbt, add the plugin to plugins.sbt, and add the snapshot resolver to both.

mkurz:
I have this exact line inside `build.sbt` and it worked for me.
@johdah Probably it doesn't work because when you only modify your `build.sbt` it doesn't re-compile your views. You have to make a manual change to your view(s) and save them so a recompile gets triggered. Or you just run `activator clean`.

mkurz:
Is it recommended to switch all view to the new DI style?
We have a controller which has quite a lot of views. Also we use constructor injection. So if we change all views to DI wouldn't this mean for each request on a specifc action ALL views gets initialized and passed to the controller even though we only need exactly one view for this request? Does it have any performance/memory implications to always, on each request, initalize alls views even though they are not needed?
I guess in future, when there is no `ThreadLocal` for the Java context anymore (https://github.com/playframework/playframework/issues/6168), we would have to pass the context via DI into the views anyway - so probably static views will not work with play anymore. (?)

johdah:
@gmethvin @mkurz  I thought I cleaned the project, but I cleaned it again and it looks like it works now. Those errors are gone, but now I have others, but those aren't related to Twirl so thanks for the help :+1:

johdah:
I have views that are used recursively for building trees. Now that causes circular dependency since they depend on themselves. If I have a view called user.scala.html that uses buildTree.scala.html. Should user pass it's injected instance of buildTree to buildTree or how should that be solved if it can be solved?

Because if I inject it into every instance of buildTree I get circular dependency errors and if I pass it as a parameter I get "Unspecified value parameter buildTree". Or can't I use views this way anymore?

gmethvin:
@mkurz It's your choice whether you want to use the DI style. If your templates are simple and stateless, static templates are probably better.

> So if we change all views to DI wouldn't this mean for each request on a specifc action ALL views gets initialized and passed to the controller even though we only need exactly one view for this request?

If your controllers are not singletons, yes. I'm not sure of the real performance implications of that. All it's doing is creating an instance of the already compiled template though.

> I guess in future, when there is no ThreadLocal for the Java context anymore (playframework/playframework#6168), we would have to pass the context via DI into the views anyway - so probably static views will not work with play anymore. (?)

That's an interesting question. If we're injecting the templates into the controller, they can't have the context injected unless we commit to always creating a new controller on each request. The other ways of doing it would probably require passing the context as a parameter to the template (though in Scala this could be an implicit parameter).

gmethvin:
@johdah I'm not sure if this is what you're referring to, but it's totally possible to write a recursive template like this:
```
@this(max: Int = 10000)
@(a: Int = 0, b: Int = 1)

@if(b < max) {
  @{a + b}
  @this(b, a + b)
}
```
Just use `@this` to reference the current template.

julienrf:
FWIW, you can alternatively use the string interpolator:

~~~ scala
import play.twirl.api.{StringInterpolation, Html}

def buildTree(…): Html =
  if (renderingLeaf) html"leaf"
  else html"tree(${buildTree(…)}, ${buildTree(…)})"
~~~

johdah:
@gmethvin Thanks, it was exactly what I wanted to achieve. I didn't know I could use "@this" for the recursive call. Now everything works, thanks for the help. Feels like a nice solution.

wsargent:
From a comment on the mailing list and pulling this together, here's how you can DI a template:

In projects/plugins.sbt:

```
addSbtPlugin("com.typesafe.sbt" % "sbt-twirl" % "1.2.0")
```

In build.sbt:

```
TwirlKeys.constructorAnnotations += "@javax.inject.Inject()"
```

In the template:

```
@this(messagesApi: play.api.i18n.MessagesApi)
@()

@{messagesApi("hello.world")}
```

In the controller:

```
public MyController @Inject()(template: views.html.index) extends Controller {
   def index = Action {
      Ok(template())
   }
}
```

eximius313:
@wsargent How the Controller would look in Java?

adis-me:
@eximius313 it is basically a java controller with DI, correct me if am wrong:

```JAVA
import play.mvc.Controller;
import views.html.homeView;
import javax.inject.Inject;

public MyController extends Controller {

    final  homeView injectedView;
 
    @Inject
    public MyController(homeView view) {
        this.injectedView = view;
    }

    public Result index() {
        return ok(injectedView.render())
    }
 
}

```

kurtjmeyoung:
How to use a template that got injected parameters inside another one?
**e.g. i have a template that looks like the following (simplified):**

```html
@views.html.error.errorpage(Messages("title.key")) {
    <div id="content">
        <p><a href="#" onclick="<doesSomething>">@Messages("accept.key")</a></p>
    </div>
}
```

**and the template "errorpage" looks something like that (simplified):**

```html
@this(config: play.Configuration)
@(title: String)(content: Html)

<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/html">
    <head>
        <title>@Messages("name") :: @title</title>
    </head>
    <body>
        <div id="page" class="container-fluid">                   
            <div id="main">
                @content
            </div>
        </div>
    </body>
</html>
```

**As you can see "errorpage" injects the play configuration. If I try to compile the project i get the follwing error:**


> 
[error] \app\views\error\error.scala.html:4: object errorpage is not a member of package views.html.error
[error] Note: class errorpage exists, but it has no companion object.
[error] @views.html.error.errorpage(Messages("title.key")) {
> 

**Am I doing something wrong or is there any way to solve this problem? thx in advance**

wsargent:
@kurtjmeyoung from https://github.com/playframework/twirl/pull/100#issuecomment-265536917 you have to pass in errorpage as a parameter:

```
@(injectedErrorPage: BaseScalaTemplate)
@injectedErrorpage(Messages("title.key")) {

}
```

PR #106, opened 2016-06-15
Ask specs to run thread-unsafe ParserTest sequentially

retronym:
We noticed in the community build that `ParserTest` intermittently fails. I think this is likely due to the tests being run in parallel and accessing the a single instance of `TwirlParser`.

https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/433/console

```
[play-twirl] [info]     + case.scala.js
[play-twirl] [error]     ! import expressions
[play-twirl] [error]      java.lang.StringIndexOutOfBoundsException: String index out of range: 161 (TwirlParser.scala:115)
[play-twirl] [error] play.twirl.parser.TwirlParser$Input.apply(TwirlParser.scala:115)
[play-twirl] [error] play.twirl.parser.TwirlParser.any(TwirlParser.scala:227)
[play-twirl] [error] play.twirl.parser.TwirlParser.anyUntil(TwirlParser.scala:242)
[play-twirl] [error] play.twirl.parser.TwirlParser.importExpression(TwirlParser.scala:382)
[play-twirl] [error] play.twirl.parser.TwirlParser.extraImports(TwirlParser.scala:845)
[play-twirl] [error] play.twirl.parser.TwirlParser.parse(TwirlParser.scala:866)
[play-twirl] [error] play.twirl.parser.test.ParserSpec$.parseTemplateString(ParserSpec.scala:44)
[play-twirl] [error] play.twirl.parser.test.ParserSpec$.$anonfun$new$12(ParserSpec.scala:81)
[play-twirl] [error] play.twirl.parser.test.ParserSpec$.$anonfun$new$11(ParserSpec.scala:81)
[play-twirl] [info] 
[play-twirl] [info]   handle string literals within parentheses
```

```
[play-twirl] [info] New twirl parser should
[play-twirl] [info]   succeed for
[play-twirl] [error]     ! static.scala.html
[play-twirl] [error]      java.lang.StringIndexOutOfBoundsException: String index out of range: 17 (TwirlParser.scala:109)
[play-twirl] [error] play.twirl.parser.TwirlParser$Input.apply(TwirlParser.scala:109)
[play-twirl] [error] play.twirl.parser.TwirlParser.single$1(TwirlParser.scala:595)
[play-twirl] [error] play.twirl.parser.TwirlParser.plain(TwirlParser.scala:605)
[play-twirl] [error] play.twirl.parser.TwirlParser.opt1$1(TwirlParser.scala:434)
[play-twirl] [error] play.twirl.parser.TwirlParser.mixed(TwirlParser.scala:466)
[play-twirl] [error] play.twirl.parser.TwirlParser.templateContent(TwirlParser.scala:821)
[play-twirl] [error] play.twirl.parser.TwirlParser.parse(TwirlParser.scala:879)
[play-twirl] [error] play.twirl.parser.test.ParserSpec$.parseString(ParserSpec.scala:23)
[play-twirl] [error] play.twirl.parser.test.ParserSpec$.$anonfun$parseStringSuccess$1(ParserSpec.scala:30)
[play-twirl] [error] play.twirl.parser.test.ParserSpec$.parseStringSuccess(ParserSpec.scala:30)
[play-twirl] [error] play.twirl.parser.test.ParserSpec$.parseSuccess(ParserSpec.scala:27)
[play-twirl] [error] play.twirl.parser.test.ParserSpec$.$anonfun$new$3(ParserSpec.scala:58)
```

### DIFF
--- a/parser/src/test/scala/play/twirl/parser/test/ParserSpec.scala
+++ b/parser/src/test/scala/play/twirl/parser/test/ParserSpec.scala
@@ -53,6 +53,8 @@ object ParserSpec extends Specification {
     }
   }
 
+  sequential
+
   "New twirl parser" should {
 
     "succeed for" in {


schmitch:
@retronym actually this doesn't fix the error and I accidentially merged it.

What it fails is actually the pattern match:

```scala
  def debug(res: TwirlParser#ParseResult): String = {
    res match {
      case parser.Success(_, rest) => s"Success"
      case parser.Error(_, rest, errors) => s"Error"
      case _ => s"No Match: ${res.getClass}"
    }
  }
```

It will yield No Match: `No Match: class play.twirl.parser.TwirlParser$Success` guess thats a scala bug.

And if I change it to the following:
```scala
[error] /Users/schmitch/projects/schmitch/play/twirl/parser/src/test/scala/play/twirl/parser/test/ParserSpec.scala:33: object TwirlParser$Success is not a case class, nor does it have an unapply/unapplySeq member
[error]       case play.twirl.parser.TwirlParser$Success(_, rest) => s"Offset: ${rest.offset()} - Length ${rest.length_}"
```

guess I revert your change. WDYT?

gmethvin:
Well, it may still be a good idea to run sequentially, even if it doesn't fix this particular problem.

schmitch:
Eh yeah it's more than good actually, even if I can fix it by turning the `class TwirlParser` into an object:

```scala
object TwirlParser {
  val shouldParseInclusiveDot: Boolean = false
... 
```
it would fail horribly.
but if I turn it into an object the source will actually run the test just fine.

So my guess is that 
```scala
class TwirlParser(val shouldParseInclusiveDot: Boolean) {
  sealed trait ParseResult
  case class Success(template: Template, input: Input) extends ParseResult
  case class Error(template: Template, input: Input, errors: List[PosString]) extends ParseResult
}
```
calling it with `parser.Success` is broken on 2.12 and this one, too.

adriaanm:
When you match `res` against a path-dependent type like `parser.Error`, it will check that `res`'s outer pointer is actually `parser`. Could that explain why it doesn't match?

schmitch:
I actually understood that.

However since Success and Error didn't depend on anything that TwirlParser provides I would actually think that the Match would work. I mean the calling convention makes it obvious that it won't work but I guess I somehow had a Java static class in my mind that will work totally different than that.

Still the code that twirl provides in the test is ugly. But I think that it is really really hard to understand what's going on here in the first place.

PR #108, opened 2016-07-22
only use a single parser instance in tests

schmitch:
Actually currently the ParserSpec has multiple parser instances, but I suppose that this isn't needed when we do the tests in sequence.

Also this makes `parser.` calls a little bit more understandable.


schmitch:
@gmethvin should we merge that? actually I guess then we could enable Scala 2.12.0-M5 builds for twirl which would be great and after that we can do so for play-doc and work ourself up to Play 2.6 WDYT?

I would be really happy to actually test out Scala 2.12 early on.

PR #109, opened 2016-07-29
adds scalajs support

schmitch:
This PR adds ScalaJS Support.

#### Things I did
- Changed all Tests to ScalaTest
- Added a shared API Module
- Removed Specs2 Dependencies
- Added a small JavaScript Escaper (inspired by https://github.com/lihaoyi/scalatags/blob/master/scalatags/shared/src/main/scala/scalatags/Escaping.scala#L48)
- Added a small XML Escaper (https://github.com/lihaoyi/scalatags/blob/master/scalatags/shared/src/main/scala/scalatags/Escaping.scala#L48)
- Fixed all tests

#### Things that could be done after this is merged

- mimimc XML Support with browser based xml to also have a NodeSeq for SJS
- maybe add some [incremental-dom](https://github.com/google/incremental-dom) functionality which would enable change-detection. actually this would mean that we need to tokenize the Templates better which should be working for all major targets.

#### Problems

Currently this needs to include the ScalaJS Plugin to contain the ScalaJS '%%%' version for cross-compiling (https://github.com/playframework/twirl/pull/109/files#diff-fdc3abdfd754eeb24090dbd90aeec2ceR51) there is an ongoing effort to better handle this in here: https://github.com/scala-native/sbt-cross-project


build.sbt
```diff
@@ -1,65 +1,74 @@
...
+lazy val api = crossProject
+    .in(file("api"))
+    .enablePlugins(PlayLibrary, Playdoc)
+    .settings(common: _*)
+    .settings(
+      name := "twirl-api",
+      libraryDependencies ++= specs2
+    )
+    .jvmSettings(
+      // ScalaJS and Commons Lang can't work under ScalaJS (yet)
```
julienrf: Typo: “**scalaXml** and Commons Lang …”

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: scala
 script:
-  - sbt +test +publishLocal plugin/test plugin/scripted
+  - sbt +test +apiJS/test +publishLocal +apiJS/publishLocal plugin/test plugin/scripted
```
schmitch: I guess aggregate the apiJS project should also work

schmitch:
Updated ScalaTest and fixed message, etc

wsargent:
LGTM

muuki88:
This is nice. A small example on how to use this would be nice. I haven't quite figured it out yet.

schmitch:
well it's pretty simple, just add the plugin to your project `.enablePlugins(SbtTwirl)` and then create a scala html file `/demo/view.scala.html` and configure the source directories `sourceDirectories in (Compile, TwirlKeys.compileTemplates) := (unmanagedSourceDirectories in Compile).value` And call it via `demo.html.view()`

after that you can put your views everywhere, even in shared modules. and they will just run fine in scala-js.

PR #113, opened 2016-08-17
Upgrade to Interplay 1.2.0

wsargent:
Adds support for 2.12.0-M5

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ project/plugins/project/
 
 # Scala-IDE specific
 .scala_dependencies
+.idea
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ script:
   - sbt +test +publishLocal plugin/test plugin/scripted
   - jdk_switcher use oraclejdk8 && cd docs && sbt test validateDocs
 sudo: false
+jdk:
+  - oraclejdk8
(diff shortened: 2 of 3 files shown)


schmitch:
strange build error. I couldn't reproduce it locally (on jdk 8) !? 
I would guess we also need:
```
addons:
  apt:
    packages:
      - oracle-java8-installer
```

schmitch:
actually even interplay 1.1.0 fails. I mean we manually need to set `crossScalaVersions` on at least the `parser` and `compiler` since the plugin depends on it. but even that fails.
If I manually add `.settings(crossScalaVersions := List("2.11.7", "2.10.6", "2.12.0-M5"))` everywhere except the `plugin` the docs won't fail, however as soon as I raise the interplay project it fails, even when keeping `crossScalaVersions`, I even raised the doc `scalaVersion` to 2.11

wsargent:
@schmitch welp.  https://github.com/playframework/playframework/blob/master/.travis.yml#L3

schmitch:
will still fail. I tought that aswell but my hosts are at 101 and still failing :(

schmitch:
Closed via https://github.com/playframework/twirl/commit/a4e1f69f4fc49fb1242cfdc6ca0313faf61418d0

PR #114, opened 2016-08-19
Superseeds: Upgrade to Interplay 1.2.0 

schmitch:
Actually twirl can't run without implicit `scalaVersion` and `crossScalaVersion`, since interplay will  actually change crossScalaVersion to 2.11.8, 2.12.0-M5 only, but on cross compilation with the plugin the first version of compiler and parser needs to be 2.10.6, `sbt` also can't detect that and fails with strange errors inside the docs project then.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ project/plugins/project/
 
 # Scala-IDE specific
 .scala_dependencies
+.idea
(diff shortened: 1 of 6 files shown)


schmitch:
@wsargent @gmethvin it would be great, now if scala 2.12.0-RC1 is out if we may merge that and publish new versions.
Especially the Scala.JS build is needed by my company.

schmitch:
Be aware that we may need to release a 1.3.X since https://github.com/playframework/twirl/commit/104128a41ed3cecfd329f1721d5f7ae4cf0286d7 changed `Manifest` to `ClassTag` so without redistributing play it won't work together.

schmitch:
Upgraded to 2.12.0-RC2 and Scala.JS 0.6.13

PR #118, opened 2016-09-01
be compatible with Scala 2.12.0-RC1

SethTisue:
I believe these changes will not affect cross-compilation


schmitch:
I bet it's best to also raise scala-js then https://github.com/playframework/twirl/blob/master/project/plugins.sbt#L4?

SethTisue:
good point, yes. I've added a commit that does that.

PR #132, opened 2017-01-03
update readme with latest versions

emanresusername:


README.md
```diff
@@ -1,5 +1,4 @@
-Twirl
-=====
+# Twirl [![Latest version](https://index.scala-lang.org/playframework/twirl/twirl-api/latest.svg?color=orange)](https://index.scala-lang.org/playframework/twirl/twirl-api)
```
emanresusername: not sure if there's a more direct way to get at this as opposed to relying on the `twirl-api` version? is there a similar api for sbt plugins?

marcospereira: You can use this version because it is consistent across all twirl components here.

PR #135, opened 2017-02-21
Add twirl dependency injection docs

wsargent:
Adds doc showing how to use twirl constructor annotation

This is tied to Play when the rest of the doc is independent, but
this seems to be a case where it wouldn't be clear why you were 
doing this otherwise...

### DIFF
--- a/docs/manual/working/javaGuide/main/templates/JavaTemplatesDependencyInjection.md
+++ b/docs/manual/working/javaGuide/main/templates/JavaTemplatesDependencyInjection.md
@@ -1,22 +1,22 @@
 # Dependency Injection with Templates
 
-Twirl templates can now be created with a constructor annotation using `@this`.  The constructor annotation means that Twirl templates can be injected into templates directly and can manage their own dependencies, rather than the controller having to manage dependencies not only for itself, but also for the templates it has to render.
+Twirl templates can be generated as a class rather than a static object by declaring a constructor using a special @this(args) syntax at the top of the template. This means that Twirl templates can be injected into controllers directly and can manage their own dependencies, rather than the controller having to manage dependencies not only for itself, but also for the templates it has to render.
 
-As an example, suppose a template has a dependency on a component `TemplateRenderingComponent`, which is not used by the controller.  
+As an example, suppose a template has a dependency on a component `Summarizer`, which is not used by the controller:
 
-First, add the `@Inject` annotation to Twirl in `build.sbt`:
+```java
+public interface Summarizer {
+  /** Provide short form of string if over a certain length */
+  String summarize(String item);
+}
+
+Create a file `app/views/IndexTemplate.scala.html` using the `@this` syntax for the constructor:
 
 ```scala
-TwirlKeys.constructorAnnotations += "@javax.inject.Inject()"
-```
+@this(summarizer: Summarizer)
+@(item: String)
 
-Then create a file `indexTemplate.scala.html` using the `@this` syntax for the constructor:
-
-```scala
-@this(trc: TemplateRenderingComponent)
-@()
-
-@{trc.render(item)}
+@{summarizer.summarize(item)}
 ```
 
 And finally define the controller in Play by injecting the template in the constructor:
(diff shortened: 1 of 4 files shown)


docs/manual/working/javaGuide/main/templates/JavaTemplatesDependencyInjection.md
```diff
@@ -0,0 +1,41 @@
...
+
+As an example, suppose a template has a dependency on a component `TemplateRenderingComponent`, which is not used by the controller.  
+
+First, add the `@Inject` annotation to Twirl in `build.sbt`:
+
+```scala
+TwirlKeys.constructorAnnotations += "@javax.inject.Inject()"
+```
+
+Then create a file `indexTemplate.scala.html` using the `@this` syntax for the constructor:
```
gmethvin: I suggest we call this `IndexTemplate` since it's a class, and suggest using that convention. That would make it easy to distinguish class and object-based templates.

docs/manual/working/javaGuide/main/templates/JavaTemplatesDependencyInjection.md
```diff
@@ -0,0 +1,41 @@
+# Dependency Injection with Templates
+
+Twirl templates can now be created with a constructor annotation using `@this`.  The constructor annotation means that Twirl templates can be injected into templates directly and can manage their own dependencies, rather than the controller having to manage dependencies not only for itself, but also for the templates it has to render.
```
gmethvin: This makes it sound like `@this` is an annotation like in Java/Scala. I would say something like

> Twirl templates can be generated as a class rather than a static object by declaring a constructor using a special `@this(args)` syntax at the top of the template. This means that Twirl templates can be injected into controllers directly and can manage their own dependencies, rather than the controller having to manage dependencies not only for itself, but also for the templates it has to render.

docs/manual/working/javaGuide/main/templates/JavaTemplatesDependencyInjection.md
```diff
@@ -0,0 +1,41 @@
...
+
+As an example, suppose a template has a dependency on a component `TemplateRenderingComponent`, which is not used by the controller.  
+
+First, add the `@Inject` annotation to Twirl in `build.sbt`:
+
+```scala
+TwirlKeys.constructorAnnotations += "@javax.inject.Inject()"
```
mkurz: This is set in [PlaySettings.scala](https://github.com/playframework/playframework/blob/2.6.0-M1/framework/src/sbt-plugin/src/main/scala/play/sbt/PlaySettings.scala#L57) already - is it really necessary for the user to set it in `build.sbt` as well?

wsargent: I did not know that was there...

wsargent: @mkurz The problem I think is that Twirl can be used OUTSIDE of Play, and if you see the twirl docs, they don't assume Play is the target (outside of this PR).

docs/manual/working/scalaGuide/main/templates/ScalaTemplatesDependencyInjection.md
```diff
@@ -0,0 +1,44 @@
...
+
+Twirl templates can be generated as a class rather than a static object by declaring a constructor using a special @this(args) syntax at the top of the template. This means that Twirl templates can be injected into controllers directly and can manage their own dependencies, rather than the controller having to manage dependencies not only for itself, but also for the templates it has to render.
+
+As an example, suppose a template has a dependency on a component `Summarizer`, which is not used by the controller:
+
+```scala
+trait Summarizer {
+  /** Provide short form of string if over a certain length */
+  def summarize(item: String)
+}
```
marcospereira: Missing ``` to close code snippet.

benmccann:
@wsargent just a reminder to update this PR so we can merge it

benmccann:
I've merged this with the requested changes: https://github.com/playframework/twirl/commit/89f00b10f1fb93d1cb6e69330b6ab7694e8a93c8

PR #136, opened 2017-05-10
Add a Java compiler API to call from Maven and Gradle

benmccann:
I've needed this code for both the Maven and Gradle plugins which are written in Java. Seems better to have it live here in a common location than have to duplicate it in both plugins.


wsargent:
This looks good, but can you add a test to run the static method and check that it gives output without blowing up? Not concerned about right now, but it's good for regression purposes.

benmccann:
@wsargent I added a test. thanks again for taking a look

benmccann:
@wsargent can you take another look?

wsargent:
👍

PR #137, opened 2017-05-10
Update dependencies

marcospereira:
Main update were:

1. sbt
2. interplay
3. Scala JS

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,6 @@
-lazy val scalatest = "3.0.0"
-lazy val scala210 = "2.10.6"
-lazy val scala211 = "2.11.8"
-lazy val scala212 = "2.12.1"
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -1,7 +1,8 @@
-lazy val scalatest = "3.0.0"
-lazy val scala210 = "2.10.6"
-lazy val scala211 = "2.11.8"
-lazy val scala212 = "2.12.1"
+import interplay.ScalaVersions._
+
+val Versions = new {
```
gmethvin: Can this be
```scala
object Versions {
  val scalatest = "3.0.3"
}
```
?

marcospereira: Done.

I've let only the `scalatest` variable here. :-)

PR #140, opened 2017-06-23
Fix if without brackets

marcospereira:


gmethvin:
@marcospereira Can you update with a more complex example where it's used inside an HTML tag?

gslowikowski:
Will you release `1.3.3` soon?

PR #142, opened 2017-08-04
Don't automatically bring in scalajs with sbt-twirl

gmethvin:
I don't believe we should be forcing users to add the scalajs plugin, especially as it brings in a lot of other potentially conflicting dependencies.


schmitch:
This will fail since:
https://github.com/playframework/twirl/blob/master/sbt-twirl/src/main/scala/play/twirl/sbt/SbtTwirl.scala#L81
However the question might be how to pull the correct dependency for scala-js?

gmethvin:
Okay. I was under the impression that `%%%` would be the same as `%%` for non-scalajs projects as suggested in the comment. It seems the real issue is `isScalaJsProject`.

schmitch:
well without the sbt-scalajs there is no way to use `%%%`, however if there is any way to check if we are running on scala-js we can generate the dependency on it's own

gmethvin:
or maybe have a separate sbt plugin for only scalajs?

schmitch:
well I guess we would need to seperate it. since using https://github.com/scala-native/sbt-crossproject would still pull scala-js :/
but I think `isScalaJsProject` can be removed, which means the only problem would be `%%%`.

gmethvin:
```
    val isScalaJSProject = SettingKey[Boolean]("isScalaJSProject",
        "Tests whether the current project is a Scala.js project. " +
        "Do not set the value of this setting (only use it as read-only).",
        BSetting)
```

so I guess we should not be setting it anyway...

schmitch:
well yeah I have no idea why I added that.
But now it fails with: `value %%% is not a member of String` that's why I added addSbtPlugin `scala-js`.
I don't think it's possible to fix that without splitting it up somehow.

gmethvin:
I wonder if we can just manually add the `_sjs0.6` manually like we suggested, if we can inspect the project and detect that it's scalajs

schmitch:
yeah...
what I do wonder is, if we define:
```
    val isScalaJSProject = SettingKey[Boolean]("isScalaJSProject",
        "Tests whether the current project is a Scala.js project. " +
        "Do not set the value of this setting (only use it as read-only).",
        BSetting)
```
and never set it and just call:
```
libraryDependencies += {
if ((isScalaJSProject ?? false).value) 
"com.typesafe.play" %% "twirl-api_sjs0.6" % twirlVersion.value
else
"com.typesafe.play" %% "twirl-api" % twirlVersion.value
}
```
the question would be, if sbt would resolve the isScalaJSProject correctly if it does run under scala-js or not.

schmitch:
well on newer scala-js version (i.e. on their master branch) they added ` crossPlatform`, which would mean that we could add `addSbtPlugin("org.scala-native" % "sbt-crossproject"         % "0.2.0")`
and query it via:
```
if (crossPlatform.value.identifier == "jvm") false
else if (crossPlatform.value.identifier == "scala-js") true
else throw new IllegalStateException("unsupported platform")
```
unfortunatly that is not inside the 0.6.x branch

schmitch:
However adding:
```
  // Since we don't want this to work even if the scalajs plugin is disabled
   private val isScalaJSProject = SettingKey[Boolean]("isScalaJSProject",
     "Tests whether the current project is a Scala.js project. " +
     "Do not set the value of this setting (only use it as read-only).",
     BSetting)
```

works, when adding the correct import `import sbt.KeyRanks.BSetting`.

i.e. setting:
```
import sbt.KeyRanks.BSetting

   private val isScalaJSProject = SettingKey[Boolean]("isScalaJSProject",
     "Tests whether the current project is a Scala.js project. " +
     "Do not set the value of this setting (only use it as read-only).",
     BSetting)

    libraryDependencies += {
       if ((isScalaJSProject ?? false).value) "com.typesafe.play" %% "twirl-api_sjs0.6"  % twirlVersion.value
       else  "com.typesafe.play" %% "twirl-api"  % twirlVersion.value
    }
```

will actually correctly resolve the correct twirl version

sbt-twirl/src/main/scala/play/twirl/sbt/SbtTwirl.scala
```diff
@@ -73,12 +71,20 @@ object SbtTwirl extends AutoPlugin {
...
+    "Do not set the value of this setting (only use it as read-only).",
+    BSetting)
+
   def dependencySettings = Def.settings(
-    // Task is undefined if not a ScalaJSPlugin
-    isScalaJSProject := (isScalaJSProject ?? false).value,
     twirlVersion := readResourceProperty("twirl.version.properties", "twirl.api.version"),
     // %%% will be the same as %% for normal projects
-    libraryDependencies += "com.typesafe.play" %%% "twirl-api" % twirlVersion.value
+    if ((isScalaJSProject ?? false).value)
```
schmitch: still misses libraryDependencies += {}

sbt-twirl/src/main/scala/play/twirl/sbt/SbtTwirl.scala
```diff
@@ -73,12 +71,20 @@ object SbtTwirl extends AutoPlugin {
...
+  private val isScalaJSProject = SettingKey[Boolean]("isScalaJSProject",
+    "Tests whether the current project is a Scala.js project. " +
+    "Do not set the value of this setting (only use it as read-only).",
+    BSetting)
+
   def dependencySettings = Def.settings(
-    // Task is undefined if not a ScalaJSPlugin
-    isScalaJSProject := (isScalaJSProject ?? false).value,
     twirlVersion := readResourceProperty("twirl.version.properties", "twirl.api.version"),
     // %%% will be the same as %% for normal projects
```
schmitch: comment can be removed

sbt-twirl/src/main/scala/play/twirl/sbt/SbtTwirl.scala
```diff
@@ -73,12 +71,20 @@ object SbtTwirl extends AutoPlugin {
...
-    libraryDependencies += "com.typesafe.play" %%% "twirl-api" % twirlVersion.value
+    if ((isScalaJSProject ?? false).value)
+      "com.typesafe.play" %% "twirl-api_sjs" % twirlVersion.value
```
schmitch: needs to be twirl-api_sjs0.6

gmethvin:
Good call. Trying out sbt-crossproject.

gmethvin:
@schmitch I think actually sbt-project solves the issuet. Can you review again?

schmitch:
well sadly it can't be seen, however `sbt-crossproject` does only work for scala-js 1.0.0-M1, scala-js would fail to resolve the twirl project.

gmethvin:
@schmitch okay, you're right. I added an actual test with a scalajs build, and it seems to be working using a hack to determine if the project is scalajs. I also discussed on scala-js/scala-js#2987. Let me know if you think there's anything else in particular I need to test and I can add it.

richdougherty:
I see a scripted test to check that the scalajs plugin dependency works OK if scalajs is present. Is it possible to add a scripted test checking that the scalajs plugin _isn't_ brought in as a dependency unless it's already present?

gmethvin:
I added a check to see if the ScalaJSPlugin class is present on the non-scalajs compile test.

I'm a little bit unsure about the scalajs scripted test because it seems to have failed on travis for an unexplained reason before I tried again and it ran successfully. But I'm unable to get it to fail on my machine.

sbt-twirl/src/main/scala/play/twirl/sbt/SbtTwirl.scala
```diff
@@ -74,11 +73,16 @@ object SbtTwirl extends AutoPlugin {
...
     twirlVersion := readResourceProperty("twirl.version.properties", "twirl.api.version"),
-    // %%% will be the same as %% for normal projects
-    libraryDependencies += "com.typesafe.play" %%% "twirl-api" % twirlVersion.value
+    libraryDependencies += {
+      val crossVer = crossVersion.value
+      val isScalaJS = CrossVersion(crossVer, scalaVersion.value, scalaBinaryVersion.value) match {
+        case Some(f) => f("").contains("_sjs0.6") // detect ScalaJS CrossVersion
+        case None => false
+      }
+      val baseModuleID = "com.typesafe.play" %% "twirl-api" % twirlVersion.value
```
schmitch: I think for forward compat, we should still add `sbt-crossproject` and use `%%%` here.
I think sooner or later we want to publish for scala-js 1.0 anyway.

schmitch:
Actually I tested it against a demo project and it looks working. It correctly pulled the sjs library.

sbt-twirl/src/main/scala/play/twirl/sbt/SbtTwirl.scala
gmethvin: Don't we need to update twirl to support scalajs 1.0 anyway?

schmitch: I didn't checked it, but besides adding the build target we are probably ready for scala-js 1.0

gmethvin: right, but I'd rather wait for another PR to add the cross building for 0.6/1.0. I can add a comment to switch when we move to 1.0 though.

schmitch: well fine for me, pr is 100% ready to merge then.

PR #147, opened 2017-09-13
Add Apache License

gmethvin:
Uses the actual apache license instead of the file header for the LICENSE file.


mkurz:
Merged, no need to wait for Travis to finish, it's just the LICENSE file.

PR #148, opened 2017-09-22
Add toString method in template position

marcospereira:
This would be useful for Play when testing routes compiler.


marcospereira:
Apparently, the version was not updated in the last release. The last release is 1.3.7 and the version file is still 1.3.7-SNAPSHOT.

I will investigate the possible cause. For a while, we can update the version manually.

PR #159, opened 2017-11-24
Change content type for JavaScript to application/json

samupra:


samupra:
@gmethvin  Please let me know if there's something more I can do :)

gmethvin:
@playframework/core I created a new branch for twirl 1.3.x since this is a potentially breaking change. We want the new Content-Type for json to be added in Play 2.7 / twirl 1.4.0.

gpgekko:
Something went wrong here, since this change ended up in Play 2.6.12 / Twirl 1.3.14.
What is the reasoning behind changing this? It now differs from the definition of the Javascript content type used in the rest of Play, which seems rather odd.

richdougherty:
Oops, it does look like (a) this change was included in the 1.3.x branch and (b) the 1.3.14 release was made from the master branch anyway.

Cc @gmethvin @marcospereira.

BTW I recently (#172) fixed up the 1.3.x branch to include all the 1.3.14 code, but this commit (ae99b3b) was already part of the 1.3.x branch prior to that PR.

marcospereira:
@richdougherty I think it was added to 1.3.x unintentionally and should be reverted to avoid problems like @gpgekko described.

I will submit a PR reverting the commit, and the next release of Twirl should be made from 1.3.x branch to avoid new problems.

WDYT?

marcospereira:
See #173.

richdougherty:
@marcospereira I've now merged that change.

PR #167, opened 2018-02-15
Add support to Scala 2.13.0-M3

marcospereira:
Keeping up with Scala 2.13 releases.

See https://github.com/playframework/playframework/issues/7940.


richdougherty:
Backported here: #172

Issue #141, opened 2017-06-30
Consider indentation and remove newlines generated for "empty" lines


Lasering:
Given the template declared in `indentation.scala.html`:

```
@(list: List[String])

<!DOCTYPE html>
<html lang="en">
  <head></head>
  <body>
    @if(list.isEmpty) {
      <h1>Nothing to display</h1>
    } else {
      <h1>@list.size items!</h1>
    }
  </body>
</html>
```

The generated html will be:
```


<!DOCTYPE html>
<html lang="en">
  <head></head>
  <body>
    
      <h1>2 items!</h1>
    
  </body>
</html>
```

Notice that the `h1` tag as an extra indentation level and it is surrounded by empty lines. This is happening because of the `@if`. Twirl could be smarter (and way more awesome) if it would remove a level of indentation from the branches of the if.

It could also remove the generated empty lines the if causes. Please note that if the if was declared like this:
`@if(list.isEmpty) { <h1>Nothing to display</h1> } else { <h1>@list.size items!</h1> }` no newlines are created.

marcospereira:
We could possibly trim the empty lines (a pull request would be good if you have the time).

But I don't think we should touch indentation:

1. It is not really super useful
2. Should we consider the indentation tabs or spaces? How many spaces?
3. What if the user had not added a level of indentation?

In other words, too many things to consider and not much value added, IMO.

helllamer:
Just strip out all useless spaces during play `stage` including identation is enought.

For example, i'm using `grunt` with `replace` task against twirl templates for minification during `stage` (but buggy `play-yeoman` is a *real pain*). This is parts from fat `Gruntfile.js`:
```
...
grunt.initConfig({
  //...
  replace: {
    txtmin: {
	src: ['<%= yeoman.dist %>/**/*.scala.txt'],
	overwrite: true,
	replacements: [
	  {from: /@\*.*?\*@/mg, to: ''},    // strip twirl comments.
	  {from: /^[ \t]+/mg, to: ''}	    // strip spaced BOL offsets
	]
      },
      
      csstplmin: {
	src: ['<%= yeoman.dist %>/**/*Css.scala.txt'],
	overwrite: true,
	replacements: [
	  {from: /^\s+/mg, to: ''},	    // BOL offsets, empty lines
	  {from: /}\s+}/mg, to: '}}'}
	]
      },

      htmlmin: {
	src: ['<%= yeoman.dist %>/**/*.scala.html'],
	overwrite: true,
[...]

PR #169, opened 2018-02-20
newLine at end of @this(...) and @(...) are not part of content

mkurz:
This is quite a simple fix and I am wondering why this wasn't addressed yet.

If you have a template that defines a constructor or takes arguments you will always have a newLine in front of your rendered content. People are complaining about this behaviour e.g [here](https://github.com/playframework/twirl/issues/141#issuecomment-329448988) and I have seen other comments where people complain about the newLine behaviour of various twirl  expressions. This PR fixes a (small) part of that complains.

E.g. given following template:
```
@(someVar:String)
This is my content
```
currently renders to the following two lines
```

This is my content
```

But actually it should be just one line, without the new line
```
This is my content
```
The same happens when you use `this(...)`

IMHO the newLine at the end of `this(...)` and `@()` should not be seens as part of the template. These two lines just define the constructor and the arguments for the render method, but have nothing to do with the content.

Eventually my fix will also remove a lot of newLines for many people because usually you template call other templates that call other templates ... etc. so you will end up with many ugly new lines no one needs or wants.

All my pr does is calling `check("\n")` after the argument lists which swallows a new line if there is one.

### DIFF
--- a/parser/src/main/scala/play/twirl/parser/TwirlParser.scala
+++ b/parser/src/main/scala/play/twirl/parser/TwirlParser.scala
@@ -902,8 +902,13 @@ class TwirlParser(val shouldParseInclusiveDot: Boolean) {
       input.regress(1)
       val p = input.offset()
       val args = templateArgs()
-      if (args != null) Some(position(PosString(args), p))
-      else None
+      if (args != null) {
+        val result = position(PosString(args), p)
+        check("\n")
+        Some(result)
+      } else {
+        None
+      }
     } else None
   }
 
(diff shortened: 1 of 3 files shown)


mkurz:
@richdougherty So you are ok to merge this? Or want someone else to approve?

marcospereira:
Thank you, @mkurz.

marcospereira:
Backport to 1.3.x: a351758

mkurz:
Great, thanks for backporting :smile:

PR #171, opened 2018-02-28
Update version - 1.3.14 was tagged on master

mkurz:
Makes sure the next release off the 1.3.x branch has the correct version.


richdougherty:
Closing this because it is now part of #172 which also includes all the other missing commits that are needed. Thanks @mkurz for making this update.

PR #180, opened 2018-07-05
Scala 2.13.0-M4

xuwei-k:


schmitch:
well I think we can also change the value inside the build.sbt and travis-ci

xuwei-k:
What do you mean? I have changed `build.sbt`

https://github.com/playframework/twirl/blob/ec2cafe1b6c959fd74b5ffd796260ce07198eb8b/.travis.yml#L5-L9

schmitch:
well nvm, but you could try to use the newest interplay which already has the scala version defined.

PR #183, opened 2018-07-25
Added security info to template escaping

stuartgunter:
A very brief explanation of template escaping is given (with code sample), but there is currently no warning of the security implications of using this technique. Given the potential for harm, I think it would be helpful for developers to understand the risks in using this.

I've added a short description as a starting point for discussion and would be grateful to receive feedback on it. I considered including links to more detailed content on XSS but wasn't sure if that fits with the documentation guidelines for twirl. Happy to add some links if anyone thinks that would be beneficial.

### DIFF
--- a/docs/manual/working/javaGuide/main/templates/JavaTemplates.md
+++ b/docs/manual/working/javaGuide/main/templates/JavaTemplates.md
@@ -242,7 +242,7 @@ You can put a comment on the first line to document your template into the Scala
 
 ## Escaping
 
-By default, dynamic content parts are escaped according to the template type’s (e.g. HTML or XML) rules. If you want to output a raw content fragment, wrap it in the template content type.
+By default, dynamic content parts are escaped according to the template type’s (e.g. HTML or XML) rules. If you want to output a raw content fragment, wrap it in the template content type. When using this feature, consider the security implications of outputting raw HTML if there is any possibility that the user can control the content. This technique is a common cause of Cross Site Scripting (XSS) vulnerabilities and is dangerous if not used with caution.
 
 For example to output raw HTML:
 
--- a/docs/manual/working/javaGuide/main/templates/JavaTemplates.md
+++ b/docs/manual/working/javaGuide/main/templates/JavaTemplates.md
@@ -242,7 +242,9 @@ You can put a comment on the first line to document your template into the Scala
 
 ## Escaping
 
-By default, dynamic content parts are escaped according to the template type’s (e.g. HTML or XML) rules. If you want to output a raw content fragment, wrap it in the template content type. When using this feature, consider the security implications of outputting raw HTML if there is any possibility that the user can control the content. This technique is a common cause of Cross Site Scripting (XSS) vulnerabilities and is dangerous if not used with caution.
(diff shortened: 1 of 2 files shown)


docs/manual/working/javaGuide/main/templates/JavaTemplates.md
```diff
@@ -242,7 +242,7 @@ You can put a comment on the first line to document your template into the Scala
 
 ## Escaping
 
-By default, dynamic content parts are escaped according to the template type’s (e.g. HTML or XML) rules. If you want to output a raw content fragment, wrap it in the template content type.
+By default, dynamic content parts are escaped according to the template type’s (e.g. HTML or XML) rules. If you want to output a raw content fragment, wrap it in the template content type. When using this feature, consider the security implications of outputting raw HTML if there is any possibility that the user can control the content. This technique is a common cause of Cross Site Scripting (XSS) vulnerabilities and is dangerous if not used with caution.
```
marcospereira: You can use a `> **Note:**` here so that the content will be highlighted (see an [example here](https://www.playframework.com/documentation/2.6.x/JavaTemplates#A-type-safe-template-engine-based-on-Scala)):

```markdown
> **Note:** When using this feature, consider...
```
Same for Scala version.

stuartgunter:
Thanks @marcospereira. I'm really impressed by the fast turnaround!

Do you know when the website will be updated to reflect the changes? I don't mean to rush you, just wanted to check it won't need to wait for a new release of the library.

gmethvin:
Since master is Twirl 1.4.x, which is used in Play 2.7.x, I'd expect this change to be at https://www.playframework.com/documentation/2.7.x/JavaTemplates (since this hasn't been backported), but it's possible the Lightbend team is doing things differently for Play 2.7.

stuartgunter:
Hmm... it feels like the docs should be separated from the main library, as this particular issue affects all versions already in use. I think users would benefit from having up-to-date docs without having to wait for a new release.

I appreciate that may not be possible now, but recommend the docs either be separated from the library or the doc publishing process be made independent of the code publishing process.

What do you think?

marcospereira:
> Since master is Twirl 1.4.x, which is used in Play 2.7.x, I'd expect this change to be at https://www.playframework.com/documentation/2.7.x/JavaTemplates (since this hasn't been backported), but it's possible the Lightbend team is doing things differently for Play 2.7.

This is how it works. We can backport this change without problems and it will later appear in /2.6.x/ docs.

> Hmm... it feels like the docs should be separated from the main library, as this particular issue affects all versions already in use.

Having code and docs together has its advantages too. For example, it is easier to maintain them in sync and also have versioned docs.

> I think users would benefit from having up-to-date docs without having to wait for a new release.

For the Play main documentation, we don't need a release to get new additions/fixes at the docs. I'm don't recall if this is true for Twirl and other standalone libraries, but I don't think so. To be clear, the docs publishing process is already independent, we just need to ensure it gets the new docs from the correct branch when there are changes.

Best.

marcospereira:
Backport to 1.3.x: e8bd5bfcd67336503df0aeccaa63bc352be198c5

PR #188, opened 2018-08-20
Optimize escape functions

isaacl:
- Use java.lang.StringBuilder instead of scala wrapper
- Rewrite Html.buildString to use bulk copy methods
  when possible.


lightbend-cla-validator:
Hi @isaacl,

Thank you for your contribution! We really value the time you've taken to put this together.

Before we proceed with reviewing this pull request, please sign the Lightbend Contributors License Agreement:

[http://www.lightbend.com/contribute/cla](http://www.lightbend.com/contribute/cla)

isaacl:
cla signed

isaacl:
ping... I also posted in chat and got no response....

marcospereira:
Hi @isaacl,

I think it would be good to have JHM benchmarks added to the project so that we can better see/prove/check performance numbers when doing such optimizations:

https://github.com/ktoso/sbt-jmh

Basically, you write a benchmark for the function and run it before and after the optimization to see how the code change affected performance.

Do you have the time to do it?

Best.

isaacl:
if you're willing to do it, that would be easier for me, but I will get to it eventually.

isaacl:
Do you have a benchmark framework to use or are you asking for simple jmh?

PR #189, opened 2018-09-03
Fix triggered execution, fixes #186

### LINKED ISSUE #186: Triggered execution for Twirl templates is broken
`project/plugins.sbt`:
```scala
addSbtPlugin("com.typesafe.sbt" % "sbt-twirl" % "1.3.15")
```

`project/build.properties`:
```properties
sbt.version = 1.2.1
```

`build.sbt`:
```scala
name := "TwirlTest"
version := "0.1"
scalaVersion := "2.12.6"
lazy val root = (project in file(".")).enablePlugins(SbtTwirl)
```

`src/main/twirl/index.scala.html`
```
@()

<html>
  <body>
  </body>
</html>
```

In sbt shell `show watchSources` (just showing relevant sources):
```
[info] * Source(
[info]   base = /home/simon/TwirlTest/src/main/twirl/index.scala.html,
[info]   includeFilter = sbt.io.AllPassFilter$@554d040d,
[info]   excludeFilter = sbt.io.NothingFilter$@3fe24670,
[info]   recursive = true,
[info] )
```

In sbt shell `~compile` then modify the `index.scala.html` in any way (eg. add more html, remove html, add parameters, remove parameters, interpolate parameters in html, etc) a compile will never be triggered.

I can test with more Sbt versions and/or Scala versions if that helps.

jrudolph:
Fixes #186.

### DIFF
--- a/sbt-twirl/src/main/scala-sbt-0.13/play/twirl/sbt/SbtTwirlCompat.scala
+++ b/sbt-twirl/src/main/scala-sbt-0.13/play/twirl/sbt/SbtTwirlCompat.scala
@@ -1,0 +1,14 @@
+/*
+ * Copyright (C) 2009-2014 Typesafe Inc. <http://www.typesafe.com>
+ */
+package play.twirl.sbt
+
+import sbt._
+import sbt.Keys._
+import play.twirl.sbt.Import.TwirlKeys._
+
+object SbtTwirlCompat {
+  def watchSourcesSettings:  Seq[Setting[_]] = Seq(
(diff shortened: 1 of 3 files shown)


ignasi35:
> We need twirl to be compatible with sbt 1 and 0.13.

Is there any major reason why that's still the case?

marcospereira:
> Is there any major reason why that's still the case?



Play supporting both versions is the main reason. It means users can have Play 2.6.x applications using sbt 0.13, and then twirl must follow. And we are not dropping sbt 0.13 for Play 2.7.0.

dwijnand:
This needs backporting, for Play 2.6.

PR #190, opened 2018-10-25
Add Scala 2.130-M5 support

marcospereira:
This also maintains the builds for M3 and M4.

/cc @SethTisue 

## Reference

See a discussion about why we need to override `display` instead of `printMessage`: https://github.com/scalate/scalate/issues/199

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,11 +1,9 @@
 import interplay.ScalaVersions._
 import sbtcrossproject.crossProject
 
-val scalatest = "3.0.6-SNAP1"
-
 val commonSettings = Seq(
   scalaVersion := scala210,
-  crossScalaVersions := Seq(scalaVersion.value, scala211, scala212, scala213)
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -1,11 +1,9 @@
 import interplay.ScalaVersions._
 import sbtcrossproject.crossProject
 
-val scalatest = "3.0.6-SNAP1"
-
 val commonSettings = Seq(
   scalaVersion := scala210,
-  crossScalaVersions := Seq(scalaVersion.value, scala211, scala212, scala213)
+  crossScalaVersions := Seq(scalaVersion.value, scala211, scala212, "2.13.0-M3", "2.13.0-M4", "2.13.0-M5")
```
marcospereira: M3 because that is what we have for Play now.

SethTisue:
> This also maintains the builds for M3 and M4

feel free to drop those any time. almost nobody bothers supporting more than one Scala milestone at a time. (but I do see your comment above about Play still being on M3. yes, that's a reasonable reason to leave M3 support in place for now, especially if doing so isn't troublesome)

I've reviewed the diffs, and I don't see a need to do a community build run before merging this.

PR #194, opened 2018-11-13
Match statement not allowed to have code block args

mkurz:
The code block of a `match` statement never has block arguments, according to the [Scala specification](https://www.scala-lang.org/files/archive/spec/2.12/13-syntax-summary.html):

```
Expr1             ::=  ... | PostfixExpr ‘match’ ‘{’ CaseClauses ‘}’
...
CaseClauses       ::=  CaseClause { CaseClause }
...
CaseClause        ::=  ‘case’ Pattern [Guard] ‘=>’ Block
```

This finally allows us to have `match ... case` statements on the same line within twirl templates :wink:

### DIFF
--- a/parser/src/main/scala/play/twirl/parser/TwirlParser.scala
+++ b/parser/src/main/scala/play/twirl/parser/TwirlParser.scala
@@ -504,12 +504,12 @@ class TwirlParser(val shouldParseInclusiveDot: Boolean) {
     }
   }
 
-  def block(): Block = {
+  def block(blockArgsAllowed: Boolean): Block = {
     var result: Block = null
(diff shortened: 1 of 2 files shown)


mkurz:
My point is (and what the Scala specs confirms):
After the opening curly braces of `match` there **has to be** the word `case`:
```
match { case ...
```
Therefore we do not even have to try to parse block args for the `match { ... }` block (by looking for `=>`). It is just not allowed to have code like:
```
match { => args ....
```

mkurz:
Thanks @renatocaval!

renatocaval:
Thank you!

PR #195, opened 2018-11-13
Remove old scala 2.13 versions

renatocaval:
This PR updates Twirl to use interplay 2.0.4 and it removes old Scala 2.13 versions. 

We are not building Play for any of the Scala 2.13 versions so in principle there is no reason for keeping them around. 

I'm keeping 2.13.0-M5 though.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ import org.scalajs.jsenv.nodejs.NodeJSEnv
 
 val commonSettings = Seq(
   scalaVersion := scala210,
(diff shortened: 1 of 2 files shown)


renatocaval:
Well, I'm in doubt now if that's the right thing to do. 

From Play perspective, it doesn't change, because we don't have a 2.13 build anyway. However, others depending on Twirl, but not on Play may be interested. 

Eventually, we should add 2.13.0-M3 because it is the one before the big collection refactoring.

PR #196, opened 2018-11-27
Fix args block parsing when there is a case statement

marcospereira:


compiler/src/test/resources/blockWithTuple.scala.html
```diff
@@ -0,0 +1,5 @@
+@(args: Seq[(String, String)])
+
+@args.map { case (k, v) =>
```
marcospereira: A small regression was causing such blocks not to be parsed correctly.

parser/src/main/scala/play/twirl/parser/TwirlParser.scala
```diff
@@ -491,8 +491,15 @@ class TwirlParser(val shouldParseInclusiveDot: Boolean) {
 
   def blockArgs(): PosString = {
     def noCurlyBraces(result: String): Boolean = !result.contains("{") && !result.contains("}")
-    def noOpeningParenthesis(result: String): Boolean = !result.contains("(") ||
-      (result.trim.startsWith("(") && result.stripSuffix("=>").trim.endsWith(")") && result.count(_ == '(') == 1)
+    def noOpeningParenthesis(result: String): Boolean = {
+      val noContainsOpenParenthesis = !result.contains("(")
+
+      val startsWithParenthesis = result.trim.startsWith("(") || result.trim.stripPrefix("case").trim.startsWith("(")
```
marcospereira: A block args can also start with a `case` keyword.

marcospereira:
Thanks for reviewing, @mkurz!

PR #213, opened 2019-04-08
Scala 2.13.0-RC1

marcospereira:
Backport won't be clean since here we are removing Scala 2.11 and we should keep it in branch 1.4.x.


marcospereira:
/cc @SethTisue :-)

The current block is that [Scala.js has support for Scala 2.13.0-RC1](https://mvnrepository.com/artifact/org.scala-js/scalajs-library) in version 1.0.0-M7, but [Scalatest does not support this version of Scala.js](https://mvnrepository.com/artifact/org.scalatest/scalatest).

SethTisue:
yeah, the problem is upstream at https://github.com/rickynils/scalacheck/issues/464

I'd suggest that for RC1, it's fine to just omit sjs 1.0, just publish for JVM and sjs 0.6, hopefully it will get straightened out in time for 2.13.0 final

dwijnand:
1.4 https://github.com/playframework/twirl/pull/214

marcospereira:
@playframework/core & @SethTisue Twirl 1.4.1 released with Scala 2.13.0-RC1 support: https://repo1.maven.org/maven2/com/typesafe/play/twirl-api_2.13.0-RC1/

PR #291, opened 2019-12-16
Fixing Travis scripts

renatocaval:


dwijnand:
Don't `set -e`.  Trying to find how to source `travis_terminate`.

ignasi35:
@renatocaval this is only a format away from complete :-)

renatocaval:
> Don't set -e

@dwijnand, can you explain why? That's how we do in most of our builds.

renatocaval:
If that's the reason http://mywiki.wooledge.org/BashFAQ/105, I still think for our case it's much simpler to use `set -e` then adding `|| exit 1` everywhere. 

Do you have any other reason for that?

renatocaval:
Superseded by #293

PR #293, opened 2019-12-17
Fix Travis builds (using stages)

renatocaval:


.travis.yml
```diff
@@ -1,26 +1,67 @@
...
+      name: "Validate documentation"
 
 cache:
   directories:
-  - "$HOME/.ivy2/cache"
-  - "$HOME/.jabba/jdk"
-  - "$HOME/.sbt"
+    - "$HOME/.coursier/cache"
+    - "$HOME/.ivy2/cache"
+    - "$HOME/.ivy2/local"
```
dwijnand: Why are you adding this?

marcospereira:
Why use stages here? It is way more verbose than the build matrix. I would prefer to remove the bash scripts instead since they aren't adding much value here. I suggest remove them and have:

```yaml
script:
  - sbt ++TRAVIS_SCALA_VERSION test scalafmtCheckAll ...
```

That way, we won't bother with `travis_terminate` as well.

dwijnand:
> That way, we won't bother with `travis_terminate` as well.

Only if it's a oneliner.

.travis.yml
renatocaval: Do you mean the `.ivy2/local`? 
I was copying it from another build and forget the remove it.

dwijnand: Please remove it.  In Play we micro-manage it elsewhere, which we don't do here, so it shouldn't be cached.

renatocaval:
> Why use stages here?

For a few reasons, first, the three scripts are being run all together, one after the other, on a single build. This gives us less visibility on what is breaking or not. 

Second, adding the validation on the first stage is interesting because we can do a quick sanity check before starting running more heavy jobs. We are doing so for Play and Lagom and is paying off. 

I agree that we don't need the scripts and we could inline them. Actually, I realise now that the reason why we had this `travis_terminate` thing was that we were running all three scripts in one single job. 

Another reason for using stages is that we can select some stages to run on PRs and others on merges or cron (although there is no cron for twirl)

Now, I personally don't like the matrix in Travis. I find stages with explicitly env variables more obvious. At least I can immediately see what is going to happen.

PR #294, opened 2019-12-18
Drop sbt 0.13 support

marcospereira:
Depends on https://github.com/playframework/interplay/pull/101


project/plugins.sbt
```diff
@@ -1,4 +1,4 @@
-addSbtPlugin("com.typesafe.play" % "interplay" % sys.props.get("interplay.version").getOrElse("2.1.4"))
+addSbtPlugin("com.typesafe.play" % "interplay" % sys.props.get("interplay.version").getOrElse("2.1.5-ee0cdca"))
```
ignasi35: If only `interplay` was using `sbt-dynver` :trollface:

build.sbt
```diff
@@ -146,9 +137,9 @@ lazy val parser = project
     commonSettings,
     mimaSettings,
     name := "twirl-parser",
-    libraryDependencies ++= scalaParserCombinators(scalaVersion.value),
-    libraryDependencies += "com.novocode"  % "junit-interface" % "0.11"           % "test",
-    libraryDependencies += "org.scalatest" %%% "scalatest"     % ScalaTestVersion % "test",
+    libraryDependencies += "org.scala-lang.modules" %% "scala-parser-combinators" % ScalaParserCombinatorsVersion % "optional",
```
ihostage, line 140: @marcospereira @mkurz 
Folks, do you remember, why scope of `scala-parser-combinators` was changed to `optional`? 🤔

mkurz: I don't think optional makes sense here anymore. It was added in f288c70. I checked out that revision, removed scala-parser-combinators and it still compiled. It seems things changed since then.
We can remove optional here I think since twirl seems to not work anymore without scala-parser-combinators.

PR #335, opened 2020-05-25
Branch merge 1.5.x into master (55871df)

renatocaval:
- fix coursier cache
- remove copyright year
- remove unnecessary .ivy2/local
- user 'exit 1' instead of 'set -e'
- using stages
- fixing travis scripts

Like #334, but sent directly from repo (not from fork)

PR #336, opened 2020-05-25
Remove interplay, prepare for release by tag

renatocaval:
Remove interplay, prepare for release by tag

Build on top of #335

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,4 +1,4 @@
-import interplay.ScalaVersions._
+import Dependencies._
 import sbtcrossproject.crossProject
 import org.scalajs.jsenv.nodejs.NodeJSEnv
 
@@ -11,39 +11,6 @@ val ScalaParserCombinatorsVersion = "1.1.2"
 
 val mimaSettings = Seq(
   mimaPreviousArtifacts := previousVersion.map(organization.value %% name.value % _).toSet
-)
-
-val javacParameters = Seq(
-  "-source",
-  "1.8",
-  "-target",
-  "1.8",
-  "-Xlint:deprecation",
-  "-Xlint:unchecked"
-)
-
-val scalacCompilerParams = Seq(
-  "-target:jvm-1.8",
-  "-Ywarn-unused:imports",
-  "-Xlint:nullary-unit",
-  "-Xlint",
-  "-Ywarn-dead-code",
-)
-
-val javaCompilerSettings = Seq(
-  javacOptions in Compile ++= javacParameters,
-  javacOptions in Test ++= javacParameters,
-)
-
-val headerSettings = Seq(
-  headerLicense := {
-    Some(
-      HeaderLicense.Custom(
-        s"Copyright (C) Lightbend Inc. <https://www.lightbend.com>"
-      )
-    )
-  },
-  headerEmptyLine := false
 )
 
 // Customise sbt-dynver's behaviour to make it work with tags which aren't v-prefixed
@@ -60,49 +27,12 @@ Global / onLoad := (Global / onLoad).value.andThen { s =>
   s
 }
 
-// this overrides interplay release settings with some adaptations for working with sbt-dynver
-// moreover, it contain only bits necessary for twirl, nothing else
-lazy val releaseSettings: Seq[Setting[_]] = Seq(
-  // Release settings
-  releaseCrossBuild := false,
-  releaseProcess := {
-    import ReleaseTransformations._
-
-    def ifDefinedAndTrue(key: SettingKey[Boolean], step: State => State): State => State = { state =>
-      Project.extract(state).getOpt(key in ThisBuild) match {
-        case Some(true) => step(state)
-        case _          => state
-      }
-    }
-
-    Seq[ReleaseStep](
-      checkSnapshotDependencies,
-      runClean,
-      releaseStepCommandAndRemaining("+test"),
-      releaseStepTask(playBuildExtraTests in thisProjectRef.value),
-      releaseStepCommandAndRemaining("+publishSigned"),
-      releaseStepTask(playBuildExtraPublish in thisProjectRef.value),
-      ifDefinedAndTrue(playBuildPromoteBintray, releaseStepTask(bintrayRelease in thisProjectRef.value)),
-      ifDefinedAndTrue(playBuildPromoteSonatype, releaseStepCommand("sonatypeBundleRelease")),
-      pushChanges
-    )
-  }
-)
-
-val commonSettings = javaCompilerSettings ++ headerSettings ++ Seq(
-  scalaVersion := scala212,
-  crossScalaVersions := Seq(scala212, scala213),
-  scalacOptions ++= scalacCompilerParams,
-)
-
 lazy val twirl = project
   .in(file("."))
-  .enablePlugins(PlayRootProject)
+  .disablePlugins(MimaPlugin)
   .settings(
-    commonSettings,
-    releaseSettings,
     crossScalaVersions := Nil, // workaround so + uses project-defined variants
-    mimaFailOnNoPrevious := false
+    publish / skip := true
   )
   .aggregate(apiJvm, apiJs, parser, compiler, plugin)
 
@@ -115,10 +45,9 @@ lazy val nodeJs = {
 
 lazy val api = crossProject(JVMPlatform, JSPlatform)
   .in(file("api"))
-  .enablePlugins(PlayLibrary, Playdoc)
+  .enablePlugins(Common, Playdoc, Omnidoc, PublishLibrary)
   .configs(Docs)
   .settings(
-    commonSettings,
     mimaSettings,
     name := "twirl-api",
     jsEnv := nodeJs,
@@ -131,9 +60,8 @@ lazy val apiJs  = api.js
 
 lazy val parser = project
   .in(file("parser"))
-  .enablePlugins(PlayLibrary)
+  .enablePlugins(Common, Omnidoc, PublishLibrary)
   .settings(
-    commonSettings,
     mimaSettings,
     name := "twirl-parser",
     libraryDependencies += "org.scala-lang.modules" %% "scala-parser-combinators" % ScalaParserCombinatorsVersion % "optional",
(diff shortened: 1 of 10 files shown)


build.sbt
```diff
@@ -60,49 +27,12 @@ Global / onLoad := (Global / onLoad).value.andThen { s =>
...
-
-val commonSettings = javaCompilerSettings ++ headerSettings ++ Seq(
-  scalaVersion := scala212,
-  crossScalaVersions := Seq(scala212, scala213),
-  scalacOptions ++= scalacCompilerParams,
-)
-
 lazy val twirl = project
   .in(file("."))
-  .enablePlugins(PlayRootProject)
```
ignasi35, line 100: I had to scratch my head on why `.enablePlugins(Common)` is not added here. But this looks good to me now.

Good job!

build.sbt
```diff
@@ -60,49 +27,12 @@ Global / onLoad := (Global / onLoad).value.andThen { s =>
...
   .settings(
-    commonSettings,
-    releaseSettings,
     crossScalaVersions := Nil, // workaround so + uses project-defined variants
-    mimaFailOnNoPrevious := false
+    mimaFailOnNoPrevious := false,
+    publish / skip := true
```
ignasi35, line 35: Wait, I think this needs an explicit `scalaVersion := Scala212` now. I think that's part of the workaround mentioned in the comment `// workaround so + uses project-defined variants` above.

build.sbt
```diff
@@ -176,26 +102,27 @@ lazy val plugin = project
...
-  (PgpKeys.publishSigned in plugin).value
-}
+// FIXME: how to port this?!
+// playBuildRepoName in ThisBuild := "twirl"
+// playBuildExtraTests := {
+//   (scripted in plugin).toTask("").value
+// }
+// playBuildExtraPublish := {
+//   (PgpKeys.publishSigned in plugin).value
+// }
```
ignasi35, line 115: This is not necessary since `pgp` signing will now happen on bintray. Not sure about the `scripted` part... looking.

build.sbt
ignasi35: I mean, it's there in the `Common` plugin ( in `override def globalSettings`) but that's not enabled here.

project/Dependencies.scala
```diff
@@ -0,0 +1,12 @@
+import sbt.Keys.libraryDependencies
+import sbt._
+
+object Dependencies {
+
+  val Scala212      = "2.12.10"
```
ignasi35, line 6:
```suggestion
  val Scala212      = "2.12.11"
```

project/Dependencies.scala
```diff
@@ -0,0 +1,12 @@
...
+  val Scala213      = "2.13.2"
+  val ScalaVersions = Seq(Scala212, Scala213)
+
+  val PlayVersion = sys.props.getOrElse("play.version", sys.env.getOrElse("PLAY_VERSION", "2.8.0"))
+
+}
```
ignasi35, line 12: I'm guessing this means `master` was not releasing `sbt-0.13` versions of the `sbt-twirl` plugin, correct?

build.sbt
```diff
@@ -60,49 +27,12 @@ Global / onLoad := (Global / onLoad).value.andThen { s =>
...
-      releaseStepCommandAndRemaining("+test"),
-      releaseStepTask(playBuildExtraTests in thisProjectRef.value),
-      releaseStepCommandAndRemaining("+publishSigned"),
-      releaseStepTask(playBuildExtraPublish in thisProjectRef.value),
-      ifDefinedAndTrue(playBuildPromoteBintray, releaseStepTask(bintrayRelease in thisProjectRef.value)),
-      ifDefinedAndTrue(playBuildPromoteSonatype, releaseStepCommand("sonatypeBundleRelease")),
-      pushChanges
-    )
-  }
-)
```
ignasi35, line 90: I think this was lost on the refactor.
Should this be on `project/Publish.scala`?

build.sbt
renatocaval: I added Common, it's needed. I will push.

build.sbt
renatocaval: removed, still checking the scripted part

project/Dependencies.scala
renatocaval: yes

ignasi35: Just saw https://github.com/playframework/twirl/pull/335#discussion_r429953809

build.sbt
renatocaval: We only publish to bintray (without promoting), we have nothing to push to git. So basically all this block can go.

ignasi35: Even the `+test`?

Ah, ofc! This will be a job in travis where tests have already run!

build.sbt
ignasi35: why is it needed?

There is no artifact to produce, there's no docs or sources in `"."`. What am I missing?

renatocaval: You are right, I added too much. 

`validateCode` is failing because this projection doesn't have headers. It previously include it, but I removed. I don't need to add the Common plugin just for it. I will solve in another way.

build.sbt
renatocaval: I don't understand why this is needed. I will check if it will work to confirm it.

build.sbt
renatocaval: @ignasi35, I did another push. 

We now have a CopyrightHeader plugin automatically added everywhere. When we call `validateCode`, that's a command alias and it will require that all projects (also aggregation projects) to have the header configure. If not, the validate fails. 

I took the path of other projects in Akka in which we add such a plugin and we don't try to disable things in aggregation projects. 

We still have another issue, the validation may is failing, but travis is ignoring it. I don't want to fix it right now. It was already so. I prefer to run manually and move this PR forward. Later I fix it. I just want to avoid a side track that can become a rabbit hole.

build.sbt
ignasi35: when using `+task` the default value is taken from the root project. Then , the fact that the root project has an empty `crossScalaVersions` also impacts how `+task` behaves.

It's a complete mess of cases and sub-cases. I learnt about all that while working on https://github.com/akka/akka-grpc/pull/907 (which is DRAFT for other reasonS.

ignasi35:
I like the split of `Header` out of `Common`. 👍🏼

build.sbt
renatocaval: for the moment, I didn't observed any issue related to that. It just works, expect that `test` fails but because of something else. It's also happening in `master` so not related with this PR.

build.sbt
```diff
@@ -60,49 +27,12 @@ Global / onLoad := (Global / onLoad).value.andThen { s =>
...
-
 lazy val twirl = project
   .in(file("."))
-  .enablePlugins(PlayRootProject)
   .settings(
-    commonSettings,
-    releaseSettings,
     crossScalaVersions := Nil, // workaround so + uses project-defined variants
-    mimaFailOnNoPrevious := false
+    mimaFailOnNoPrevious := false,
```
ennru, line 34: Please do `.disablePlugins(MimaPlugin)` instead.

project/Dependencies.scala
```diff
@@ -0,0 +1,12 @@
+import sbt.Keys.libraryDependencies
+import sbt._
+
+object Dependencies {
+
+  val Scala212      = "2.12.11"
+  val Scala213      = "2.13.2"
```
ennru, line 7: .travis.yml is still on 2.13.1

project/Dependencies.scala
```diff
@@ -0,0 +1,12 @@
...
+  val ScalaVersions = Seq(Scala212, Scala213)
+
+  val PlayVersion = sys.props.getOrElse("play.version", sys.env.getOrElse("PLAY_VERSION", "2.8.0"))
```
ennru, line 10: Is this used?

ignasi35:
@renatocaval I think there's [a setting missing](https://github.com/playframework/interplay/blob/master/src/main/scala/interplay/PlaySbtPluginBase.scala#L22) for sbt plugin projects affecting the `scripted` tests.

renatocaval:
@ennru / @ignasi35, it's now rebased with `master` and commits squashed. Also added the bintray credentials. 

I will run publish from my machine. This is ready for merge, only missing bit to review is the latest .travis.yml changes.

build.sbt
```diff
@@ -60,49 +27,12 @@ Global / onLoad := (Global / onLoad).value.andThen { s =>
...
-  .enablePlugins(PlayRootProject)
+  .disablePlugins(MimaPlugin)
   .settings(
-    commonSettings,
-    releaseSettings,
     crossScalaVersions := Nil, // workaround so + uses project-defined variants
-    mimaFailOnNoPrevious := false
+    publish / skip := true
   )
   .aggregate(apiJvm, apiJs, parser, compiler, plugin)
```
ignasi35, line 37: The `twirl` project doesn't declare (directly or via a plugin) a `scalaVersion`. It's probably fine since all projects get it declared via the `Common` plugin. It's only a bit surprising.

build.sbt
```diff
@@ -115,10 +45,9 @@ lazy val nodeJs = {
 
 lazy val api = crossProject(JVMPlatform, JSPlatform)
   .in(file("api"))
-  .enablePlugins(PlayLibrary, Playdoc)
+  .enablePlugins(Common, Playdoc, Omnidoc, PublishLibrary)
```
ignasi35, line 48: It's counterintuitive to have the `Playdoc` plugin on the `api` project and the `docs/` folder on the root folder.

It works because `Playdoc` uses a hardcoded path and piggybacks the `publish` task from `api` but `Playdoc` is completely unrelated to `api`. 

^^^ Not a comment, just a realization. 😒

renatocaval: I'm fine to remove it if it doesn't make sense. I keep it because it was explicitly added to it, as you can see from the diff.

build.sbt
renatocaval: I think it's fine since it's a aggregation project.

renatocaval:
I was able to publish from my machine to Bintray (snapshot repos)

CleanShot 2020-05-26 at 16 13 16

CleanShot 2020-05-26 at 16 12 43

PR #337, opened 2020-05-25
Fix scalafmt formatting

renatocaval:
Builds on top of #335. 

For some reason, the formatting is not being applied and the test is still passing. This fix it the formatting for now and can be merged after #335. 

I will fix the check in #336 since on that PR I'm touching the build i many places.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,7 +1,9 @@
-# Keep in sync with docs/.scalafmt.conf
-align = true
+version = 2.5.3
+
+align.preset = true
+danglingParentheses.preset = true
+
 assumeStandardLibraryStripMargin = true
-danglingParentheses = true
 docstrings = JavaDoc
 maxColumn = 120
 project.git = true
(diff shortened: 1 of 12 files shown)


renatocaval:
On #335 is merged, I can rebase here and merge. This is ready for review.

renatocaval:
Rebased with #335. Ready for review and merge

.scalafmt.conf
```diff
@@ -1,12 +1,13 @@
-# Keep in sync with docs/.scalafmt.conf
-align = true
+version = 2.5.3
+
+align.preset = true
+danglingParentheses.preset = true
```
renatocaval, line 4: just removing some warnings

scripts/validate-code.sh
```diff
@@ -2,4 +2,4 @@
 
 # Copyright (C) Lightbend Inc. <https://www.lightbend.com>
 
-sbt ";+validateCode;+mimaReportBinaryIssues" || exit 1
+sbt "validateCode;+mimaReportBinaryIssues" || exit 1
```
renatocaval, line 5: the cross build for `validateCode` was the issue. This is a command alias. 

It was emitting

> [warn] Issuing a cross building command, but not all sub projects have the same cross build configuration. This could result in subprojects cross building against Scala versions that they are not compatible with. Try issuing cross building command with tasks instead, since sbt will be able to ensure that cross building is only done using configured project and Scala version combinations that are configured.

and then ignoring the code. We haven't done any formatting for a while in this project.

PR #339, opened 2020-05-26
Fix for running with GraalVM

renatocaval:
Not such an important fix, but when running with GraalVM, tests will fail with 

```scala
[info] Passed: Total 35, Failed 0, Errors 0, Passed 35
[error] stack trace is suppressed; run last apiJS / Test / loadedTestFrameworks for the full output
[error] (apiJS / Test / loadedTestFrameworks) org.scalajs.testcommon.RPCCore$RPCException: scala.scalajs.js.JavaScriptException: TypeError: Access to host class org.scalacheck.ScalaCheckFramework is not allowed or does not exist.
```

See https://github.com/scala-js/scala-js/issues/3673

I was hit by this one, not pleasant.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -51,6 +51,15 @@ lazy val api = crossProject(JVMPlatform, JSPlatform)
     mimaSettings,
     name := "twirl-api",
     jsEnv := nodeJs,
+    // hack for GraalVM, see: https://github.com/scala-js/scala-js/issues/3673
+    // and https://github.com/playframework/twirl/pull/339
+    testFrameworks := List(
+      new TestFramework(
+        "org.scalatest.tools.Framework",
+        "org.scalatest.tools.ScalaTestFramework",
+        "com.novocode.junit.JUnitFramework"
+      )
+    ),
     libraryDependencies += "org.scala-lang.modules" %%% "scala-xml" % ScalaXmlVersion,
     libraryDependencies += "org.scalatest"          %%% "scalatest" % ScalaTestVersion % "test",
   )
--- a/build.sbt
+++ b/build.sbt
@@ -56,8 +56,7 @@ lazy val api = crossProject(JVMPlatform, JSPlatform)
     testFrameworks := List(
       new TestFramework(
         "org.scalatest.tools.Framework",
-        "org.scalatest.tools.ScalaTestFramework",
-        "com.novocode.junit.JUnitFramework"
+        "org.scalatest.tools.ScalaTestFramework"
       )
     ),
     libraryDependencies += "org.scala-lang.modules" %%% "scala-xml" % ScalaXmlVersion,


build.sbt
```diff
@@ -122,6 +122,7 @@ lazy val api = crossProject(JVMPlatform, JSPlatform)
     mimaSettings,
     name := "twirl-api",
     jsEnv := nodeJs,
+    testFrameworks := List(new TestFramework("utest.runner.Framework")),
```
raboof, line 125: aren't we actually using ScalaTest as a test framework in this project?

renatocaval: As I understood it, this trick you tell it to not lookup the other frameworks, see Sébastien comment here: https://github.com/scala-js/scala-js/issues/3673#issuecomment-497930605

On the other hand, this is only affecting local builds with GraalVM (works NOT on my machine syndrome). We will probably move to scala.js 1.1.0 soon (see #341). So maybe not worth it.

raboof: right, so since we're using ScalaTest as our framework, we should list ScalaTest here (and not load the others)?

renatocaval: yes, but what I did is wrong. Stupid copy-n-paste from the comment in scala.js issue. "utest.runner.Framework" doesn't exists.

This is the default value of the settings:

```scala
sbt:twirl> testFrameworks
[info] compiler / testFrameworks
[info] 	List(TestFramework(org.scalacheck.ScalaCheckFramework), TestFramework(org.specs2.runner.Specs2Framework, org.specs2.runner.SpecsFramework), TestFramework(org.specs.runner.SpecsFramework), TestFramework(org.scalatest.tools.Framework, org.scalatest.tools.ScalaTestFramework), TestFramework(com.novocode.junit.JUnitFramework))
```

We should fix it to `TestFramework(org.scalatest.tools.Framework, org.scalatest.tools.ScalaTestFramework)` and drop all the others.

I will fix it.

build.sbt
```diff
@@ -122,6 +122,9 @@ lazy val api = crossProject(JVMPlatform, JSPlatform)
     mimaSettings,
     name := "twirl-api",
     jsEnv := nodeJs,
+    // hack for GraalVM, see: https://github.com/scala-js/scala-js/issues/3673 
+    // and https://github.com/playframework/twirl/pull/339
+    testFrameworks := List(new TestFramework("org.scalatest.tools.Framework","org.scalatest.tools.ScalaTestFramework")),
```
renatocaval, line 127: @raboof, this is the right settings. 

I'm only changing it on `twirl-api` as it only impacts the scala.js part.

renatocaval:
@raboof, code is now formatted

I rebase with master and squashed the commits. 

Tests are running.

Issue #209, opened 2019-03-31
TwirlCompiler: Make output reproducible-friendly


BardurArantsson:
Hi,

Currently the TwirlCompiler emits a comment like this:

```
                  -- GENERATED --
                  DATE: Wed Mar 06 05:39:00 CET 2019
                  SOURCE: /home/bardur/wd/cwconsult/silo/modules/postgresql/src/main/twirl/dk/cwconsult/silo/postgresql/core/impl/LockStreamSQL.scala.txt
                  HASH: aca20403ba81a6a79ef8db751c2a0f13b68222eb
                  MATRIX: 651->1|764->22|791->23|829->35|864->50|892->51
                  LINES: 15->1|20->2|21->3|21->3|21->3|21->3
                  -- GENERATED --
```          

Unfortunately, this contains two pieces of information that are bad for reproducible builds:

- The absolute path of the source file.
- The timestamp of compilation.

(It's generally also bad for build systems where you ideally want the output of a build step to be stable if its inputs are -- to avoid e.g. spurious rebuilds.)

It would be much appreciated if they could either just be outright removed[0] or if we could have an option to disable embedding these 'fields'. (I'm not sure if the ordering on e.g. MATRIX and LINES output is deterministic, but it should be made so as well.)

If either of these seems acceptable, I think I should be able to code it up fairly quickly.

[0] I'm not really familiar with the Play ecosystem, but one assumes that this may be for template-debugging purposes where a dev server can show the file?

BardurArantsson:
Anyone?

marcospereira:
Hey @BardurArantsson, it is unclear to me what is the problem here. Are the builds relying on this information? If so, how?

BardurArantsson:
The TL;DR is that the output changes needlessly on recompilation of the same input -- which makes interacting with build systems like Pants, Bazel, etc. less than ideal because they'll assume that because the twirl output file has changed (its hash has changed), they need to recompile the output (and whatever depends on that output).

Does that make sense?

stephane303:
Any progress on this ? We now have 452 view files in our play project. Everytime we change ANY of the view scala file, ALL generated scala file get recompiled, this is critical to our productivity.

renatocaval:
@stephane303, this is not a planned feature/change. Of course, if someone wants to send a PR it's mostly welcome.

stephane303:
[...]

PR #378, opened 2020-10-20
Remove the generation date, keep the source relative

raboof:
This way the generated code is consistent between runs, even
on different machines (different absolute paths).

Refs #209

The tests are green, but I'm not entirely confident about all the machinery, so this probably needs some better testing first. /cc @ignasi35


ignasi35:
👍🏼 to keep `SOURCE` so the plugin can track back.

I wonder if `DATE` is read at all. I can't remember, TBH. I guess it was read back in the day but not anymore?

ignasi35:
Did a quick manual test and this LGTM.

ignasi35 (approved):

mkurz:
@Mergifyio backport 1.5.x

PR #393, opened 2021-02-10
version bumps and cleanups

SethTisue:
Motivation: I want everything modernized before I take a stab at adding Scala 3 support.

About the Scala.js upgrade, it's true that upgrading to 1.4.0 means users must use 1.4.0 as well, but I checked in with Seb about that recently and he says don't think about it just do it, as he has said a number of times before. It's considered normal in the Scala.js world that in order to upgrade your libraries you may need to be on latest Scala.js first.

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -1,11 +1,2 @@
 pullRequests.frequency = "@weekly"
-
-updates.pin  = [
-  // "It is not forward binary compatible with 1.0.x: libraries compiled with 1.1.0 cannot be used with 1.0.x."
-  // https://users.scala-lang.org/t/announcing-scala-js-1-1-0/6053/3?u=sjrd
-  { groupId = "org.scala-js", artifactId= "sbt-scalajs", version = "1.0." }
-  // Scalatest 3.2.x pulls in ScalaJs 1.1
-  { groupId = "org.scalatest", artifactId = "scalatest", version = "3.1." }
-]
-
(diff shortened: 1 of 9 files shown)


SethTisue:
OKAY WHO'S AROUND TO MERGE THIS BAD BOY

SethTisue:
Note that #391 and #386 should be closed.

mkurz:
Scala.js 1.5.0 released already, see #395

SethTisue:
I've added a commit with a 1.5.0 upgrade.

SethTisue:
can somebody review/merge this?

ignasi35 (approved):
LGTM

ignasi35:
Hi @SethTisue, I'm guessing this changes in scalajs means in the next release of `twirl` we should bump to `1.6.0` (we're now on `1.5.x`).

SethTisue:
@ignasi35 I don't think so. I have not seen other libraries doing that.

mkurz:
Hey, I am currently working on a couple of Play pull requests and I think Play will need a 2.9 release at some point this year anyway. I don't think Play will be able to support upcoming JDK's (in particular Java 17 LTS in September) and probably also Scala 3 in it's 2.8.x branch and at the same time stay binary compatible... So I think it makes sense to think about a Play 2.9 release... then we could include libraries with major releases (twirl, play-json, play-ws,...), drop stuff (Scala 2.12? sbt 0.13 was removed already from the master branch and interplay)

Issue #390, opened 2021-01-05
Scala 3 support plans?


rossabaker:
Are there plans to cross-publish Twirl for Scala 3?  We have an http4s integration I'd like to keep alive, and we're trying to identify the correct repo structure as we reorganize ourselves.

Thanks for the great tool!

(Bonus keyword: Dotty)

SethTisue:
is there anyone watching the repo who would like to take a stab at a PR, even just to see what issues arise? you wouldn't need to commit to seeing it through to the end, any amount of progress would be helpful

rossabaker:
I was able to keep it working for now thusly:

```scala
      libraryDependencies.value.map {
        case module if module.name == "twirl-api" =>
          module.withDottyCompat(scalaVersion.value)
        case module => module
      }
```

I don't recommend publishing libraries on `withDottyCompat`, but it seems a fair interim solution for applications.

SethTisue:
There is a WIP PR at #394 that I intend to return to... hoping for May. We're a bit backed up dealing with the Bintray situation :-/

brbrown25:
I'd be willing to pick up work on this. I should have some availability next week

mkurz:
@brbrown25 That would be great! Is there anything I can help you with?

brbrown25:
> @brbrown25 That would be great! Is there anything I can help you with?

I'll start hacking away at it and give you an update next week, just need to familiarize myself with the source as I've mostly been an end user of twirl!

mkurz:
@brbrown25 Thanks, if you need guidance please let me know.

brbrown25:
Update, making some good progress. Gonna push up my WIP draft tomorrow. I got most of the basics in place, just need to work through the compiler pieces!

brbrown25:
Update, the WIP pr has the correct structure and the most of the cross compilation works. Only thing outstanding is updating the scala 3 version of the compiler module. There's some notes in the discussion on the pr what places will be to be updated. Also the majority of warnings have been resolved except for the CollectionConverts helper

shannonbarnes:
> I was able to keep it working for now thusly:
> 
> ```scala
>       libraryDependencies.value.map {
>         case module if module.name == "twirl-api" =>
>           module.withDottyCompat(scalaVersion.value)
>         case module => module
>       }
> ```
> 
> I don't recommend publishing libraries on `withDottyCompat`, but it seems a fair interim solution for applications.

[...]

PR #394, opened 2021-02-10
add Scala 3 to crossbuild

SethTisue:
~includes commits from #393 — I'll rebase this one once that one's merged~

references #390


SethTisue:
status: this is still in early stages, I am still just chewing through bread and have yet to take a bite of meat

build.sbt
```diff
@@ -5,16 +5,16 @@ import org.scalajs.jsenv.nodejs.NodeJSEnv
 // Binary compatibility is this version
 val previousVersion: Option[String] = Some("1.5.0")
 
-val ScalaTestVersion              = "3.1.4"
-val ScalaXmlVersion               = "1.3.0"
-val ScalaParserCombinatorsVersion = "1.1.2"
+val ScalaTestVersion              = "3.2.3"
```
sullis, line 8: use ScalaTest 3.2.5

sullis (commented):
ScalaTest 3.2.5

project/Common.scala
```diff
@@ -28,9 +28,7 @@ object Common extends AutoPlugin {
     "-deprecation",
     "-feature",
     "-unchecked",
-    "-encoding",
-    "utf8",
-    "-target:jvm-1.8",
```
mkurz, line 33: Why did you remove `-target:jvm-1.8`?

SethTisue: It should have been removed at the time 2.11 support was dropped. It's the default on 2.12 and 2.13 and 3.0 regardless of what JVM you're building on.

SethTisue:
Travis-CI config seems messed up, there is Scala 3 stuff in the Scala 2 jobs...

SethTisue:
just a note that I haven't forgotten about this. I've been working on play-json instead, but this is next in line.

SethTisue:
> Travis-CI config seems messed up, there is Scala 3 stuff in the Scala 2 jobs

Fixed now. But there are still two non-Scala 3 problems.

The plugin test job fails with `Conflicting cross-version suffixes`

And also, MiMa is failing:

```
[error] twirl-parser: Failed binary compatibility check against com.typesafe.play:twirl-parser_2.12:1.5.0! Found 1 potential problems
[error]  * method several(scala.Function0,scala.collection.mutable.Buffer,scala.reflect.Manifest)scala.collection.mutable.Buffer in class play.twirl.parser.TwirlParser's type is different in current version, where it is (scala.Function0,scala.collection.mutable.Buffer,scala.reflect.ClassTag)scala.collection.mutable.Buffer instead of (scala.Function0,scala.collection.mutable.Buffer,scala.reflect.Manifest)scala.collection.mutable.Buffer
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("play.twirl.parser.TwirlParser.several")
```

SethTisue:
me up in `.travis.yml` trying to understand how `matrix:` and `stage:` interact:

tenor

SethTisue:
Man, I hate builds that have an sbt plugin subproject mixed in with crossbuilt subprojects. It makes sense for such builds to exist, but 🙀

SethTisue:
I think I got the CI matrix sorted out, including the plugin tests. I don't like omitting `plugin` from the root project aggregate, but I don't know how to improve that.

So what remains here is:

* MiMa is failing because `Manifest` vs `ClassTag`. I guess we ought to just bite the bullet here and get off `Manifest`, even if it means accepting the binary incompatibility and bumping the major version?
* the actual Scala 3 work

SethTisue:
Still chipping away at this. A lot of PRs (mostly version bumps) were merged since I initially submitted this, so I rebased. And then I made separate `scala-2` and `scala-3` source directories with just the version-dependent bits in them, but right now the Scala 3 ones are just `???` stubs.

SethTisue:
abandoning; someone else is free to pick it up

SethTisue:
some of the context for this is at https://discuss.lightbend.com/t/github-project-abandonned/8781/12, by the way

Issue #404, opened 2021-03-17
Failed +test on a fresh checkout of 1.5.x


ignasi35:
Clone the repo and checkout commit 915fddbb65719ead69ebc3b110d4b15394de54f0.

Run `sbt +test`.

Depending on the launch script vendor and version the command above will consistently fail or will fail once before it succeeds.

For example: using vanilla sbt freshly downloaded from https://www.scala-sbt.org/ the first execution of `sbt +test` fails but the second execution consistently succeeds.

ignasi35:
cc @dwijnand we could use your _fu_

SethTisue:
(Note that Dale is on break this week and next.)

mkurz:
I can reproduce (and [have a fix](405) ;): It's a race condition.

It happens only when running the tests against Scala 2.10.7.
I could reproduce _everytime_ with Java 8 (openjdk version "1.8.0_282") and sometimes, but not always, with Java 11 (openjdk version "11.0.10" 2021-01-19).
I have sbt 1.4.8 installed on my Ubuntu notebook.

I started with the `~/.sbt` folder removed, but actually I think that is not even necessary, if you want to reproduce just make sure
- you use Java 8 to constantly run into the issue and
- **everytime** after shutting down sbt remove the `target` folder(s), e.g. I did: `rm -rf target/ project/target/` (for a clean run next time you run sbt)

How to reproduce:
`git checkout 915fddbb65719ead69ebc3b110d4b15394de54f0`

Then
`sbt -java-home /path/to/adoptopenjdk8/ ++2.10.7 test`




<details>
 <summary>Click here to see the errors (because it's a race condition, sometimes different errors occur)</summary>

```
[info] welcome to sbt 1.3.13 (AdoptOpenJDK Java 1.8.0_282)
[info] loading global plugins from /home/mkurz/.sbt/1.0/plugins
[info] loading settings for project twirl-build from plugins.sbt ...
[info] loading project definition from /home/mkurz/work/twirl/project
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[info] loading settings for project twirl from build.sbt ...
[info] set current project to twirl (in build file:/home/mkurz/work/twirl/)
[info] Setting Scala version to 2.10.7 on 5 projects.
[info] Excluded 1 projects, run ++ 2.10.7 -v for more details.
[info] Reapplying settings...
[info] set current project to twirl (in build file:/home/mkurz/work/twirl/)
[...]

PR #405, opened 2021-03-17
Fix race condition when running tests against Scala 2.10.7

mkurz:
See https://www.scala-sbt.org/1.x/docs/Parallel-Execution.html#Defining+Restrictions
@ignasi35 :smile:


mkurz:
Fixes #404

ignasi35 (approved):
LGTM

PR #439, opened 2021-08-31
Downgrade scala-parser-combinators and scala-xml

mkurz:
twirl is a sbt-plugin. sbt-plugins get published for Scala 2.12.
The 2.12.x scala compiler depends on 1.x versions for both scala-parser-combinators and scala-xml, see https://github.com/scala/scala/blob/2.12.x/versions.properties#L21-L22
If we would use 2.x versions, we would run into annoying dependency conflicts in sbt 1.5+
There is no need to upgrade these versions to 2.x, the only reason would be if we would cross build for Scala 3, but sbt-plugins do not need to be cross build for Scala 3.

See discussion in https://github.com/scala/scala/pull/9743

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -1,2 +1,8 @@
 pullRequests.frequency = "@weekly"
 commits.message = "${artifactName} ${nextVersion} (was ${currentVersion})"
+
+updates.pin = [
+  // See comment in build.sbt why we want to pin these two deps to 1.x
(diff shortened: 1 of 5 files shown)


SethTisue:
add a `.scala-steward.conf` file to prevent the steward from attempting the upgrades?

mkurz:
> add a `.scala-steward.conf` file to prevent the steward from attempting the upgrades?

Fantastic idea!

mkurz:
@SethTisue Done!

SethTisue (approved):

PR #461, opened 2021-11-19
feat(scala3): WIP working on scala 3 cross compilation. #390.

brbrown25:
- [x] scala 3 rewrite compiler specific code


ahjohannessen:
@brbrown25 What is the status on this? :)

brbrown25:
> @brbrown25 What is the status on this? :)

I had something come up at work and didn't get to make progress, but next week I have time carved out and will push some updates :)

ahjohannessen:
👍

brbrown25:
@ahjohannessen I've gotten about as far as I can with this. My understanding of macros is limited so I tried to stumble through. I will say that the only thing remaining is to fix up `compiler/src/main/scala-3/play/twirl/compiler/TwirlCompiler.scala` otherwise I've validated that the others cross compile fine. The only thing we can't make cross compiled is the plugin as it looks like that stops support at 2.12. Sorry I couldn't get it any further along.

ahjohannessen:
@brbrown25 sbt build definitions are on Scala 2.12, so that is expected. I suppose what is missing is some kind soul that knows Scala 3 macros to step up :)

brbrown25:
> @brbrown25 sbt build definitions are on Scala 2.12, so that is expected. I suppose what is missing is some kind soul that knows Scala 3 macros to step up :)

Yeah I got it about as far as I could understand, it's definitely new and exciting territory!

ahjohannessen:
@brbrown25 Which macros need porting?

brbrown25:
> @brbrown25 Which macros need porting?

the following are the things of note. The big one is figuring out the correct imports/way of handling ValDef and Tree. But the things requiring fixing should be isolated to that file and the `TemplateAsFunctionCompiler` object

https://github.com/playframework/twirl/blob/8b5fe47d85a966b6dbbf458c23a45750d36bd5b5/compiler/src/main/scala-3/play/twirl/compiler/TwirlCompiler.scala#L511
https://github.com/playframework/twirl/blob/8b5fe47d85a966b6dbbf458c23a45750d36bd5b5/compiler/src/main/scala-3/play/twirl/compiler/TwirlCompiler.scala#L629
https://github.com/playframework/twirl/blob/8b5fe47d85a966b6dbbf458c23a45750d36bd5b5/compiler/src/main/scala-3/play/twirl/compiler/TwirlCompiler.scala#L657 (the updating the bootclasspath)
https://github.com/playframework/twirl/blob/8b5fe47d85a966b6dbbf458c23a45750d36bd5b5/compiler/src/main/scala-3/play/twirl/compiler/TwirlCompiler.scala#L681
https://github.com/playframework/twirl/blob/8b5fe47d85a966b6dbbf458c23a45750d36bd5b5/compiler/src/main/scala-3/play/twirl/compiler/TwirlCompiler.scala#L701

ahjohannessen:
@brbrown25 Briefly
Scanned through the code, seems there is no
macro usage as far as I can tell. Is it more about interacting with the new compiler API that is the hurdle?

brbrown25:
> @brbrown25 Briefly Scanned through the code, seems there is no macro usage as far as I can tell. Is it more about interacting with the new compiler API that is the hurdle?

Yeah the compiler Api is the hurdle, although tts actually possibly that macros could be used in place of the compiler Api, for this case, but I'm not 100% on that.

build.sbt
```diff
@@ -1,19 +1,22 @@
...
-// See https://github.com/scala/scala/pull/9743
-val ScalaParserCombinatorsVersion = "1.1.2" // Do not upgrade beyond 1.x
-val ScalaXmlVersion               = "1.3.0" // Do not upgrade beyond 1.x
+def parserCombinators(version: String) = {
+  CrossVersion.partialVersion(version) match {
+    case Some((2, _)) =>
+      Seq("org.scala-lang.modules" %% "scala-parser-combinators" % "1.1.2" % Optional)
+    case _ => Seq("org.scala-lang.modules" %% "scala-parser-combinators" % "2.1.0" % Optional)
+  }
+}
```
mkurz, line 19: I just commited a similiar solution to the `main` branch (https://github.com/playframework/twirl/commit/ab2821f82728b76accdc7a723326f2ada03d16bd), please rebase.

brbrown25: I'll do that this morning!

brbrown25: Done!

ahjohannessen:
@brbrown25 This seems interesting https://github.com/tototoshi/twirl/commit/6bcb7726f630639eb59087ef8b9ef52d51800e88

brbrown25:
> @brbrown25 This seems interesting https://github.com/tototoshi/twirl/commit/6bcb7726f630639eb59087ef8b9ef52d51800e88

@ahjohannessen I can pull in those changes and update this probably late this evening for first thing tomorrow!

tototoshi:
My branch is here. https://github.com/tototoshi/twirl/tree/scalameta

I think it might be easier to share the source code between scala2 and scala3 using scalameta than using presentation compiler. (I don't know if twirl can depend on scalameta 🤔 ).

The main code already can be built in scala3, but the test code not yet. It has tests that confirm the generated codes can be really compiled, so still requires dotty.

brbrown25:
> My branch is here. https://github.com/tototoshi/twirl/tree/scalameta
> 
> 
> 
> I think it might be easier to share the source code between scala2 and scala3 using scalameta than using presentation compiler. (I don't know if twirl can depend on scalameta 🤔 ).
> 

I'm gonna play around with that and I agree if we came remove the presentation compiler piece it should make life a lot easier.

brbrown25:
@tototoshi  I merged your work into mine and then tweaked mine to accommodate your scalameta changes, indeed it gets compiler to run the compile stage happily enough. The only piece that remains is updating the CompilerSpec so that the tests pass under all versions. I'll try and look at that today, but feel free to hack around with what's in this, I think we're getting pretty close!

tototoshi:
I have implemented a dotty-based test helper in my working branch to support Scala3. So now it can be built with Scala3, including the tests.
https://github.com/tototoshi/twirl/tree/scalameta

And I found another problem.

In Scala2, an `if` expression without `else` returns `Any` or `AnyVal`, but in Scala3, it seems to return `Unit`.
It makes Scala3 display nothing when using twirl `@if` without `else`.

```scala
// Scala 2.13
scala> val a = if (true) 1
val a: AnyVal = 1

scala> val a = if (true) "a"
val a: Any = a
```

```scala
// Scala 3
scala> val a = if (true) 1
1 warning found
-- [E129] Potential Issue Warning: ---------------------------------------------
1 |val a = if (true) 1
  |                  ^
  |A pure expression does nothing in statement position; you may be omitting necessary parentheses

longer explanation available when compiling with `-explain`

scala> println(a)
()
```


I modified the TwirlParser to interpret the `if` without `else` as Twirl syntax instead of Scala code. 
I think the implementation needs some refactoring, but It passes tests at least.

brbrown25:
> I have implemented a dotty-based test helper in my working branch to support Scala3. So now it can be built with Scala3, including the tests.
> 
> https://github.com/tototoshi/twirl/tree/scalameta

I'll merge this into mine tomorrow morning!

brbrown25:
@ahjohannessen @tototoshi @mkurz I have merged in the latest work from Totoshi's branch into mine and did a few cleanup things, but from my local testing I believe we now officially have twirl support for Scala 3! Thanks for the assistance and collaboration everyone :)

mkurz:
Nice! Thanks for the hard work! I will have a look when time allows :+1:

tototoshi:
@brbrown25  Thanks for your review!
I added a few more commits that refactor the parser for the `if` expression. Could you review that one too?

brbrown25:
> I added a few more commits that refactor the parser for the if expression. Could you review that one too?

@tototoshi I've added you as a collaborator, so you should be able to pull my branch and add your commits directly to it :) that would probably make this a little bit smoother if you don't mind.

brbrown25:
> > I added a few more commits that refactor the parser for the if expression. Could you review that one too?
> 
> @tototoshi I've added you as a collaborator, so you should be able to pull my branch and add your commits directly to it :) that would probably make this a little bit smoother if you don't mind.

I actually went ahead and merged in your commits really quick, they look good to me :)

brbrown25:
- [ ] Look into mimaReportBinaryIssues failure

tototoshi:
The mima errors are related to changes in methods in `TwirlParser` and `TwirlCompiler`. Since these classes are not directly used by users, probably the errors are not fatal.

tototoshi:
Anyway, it's true that the changes are breaking changes and switching to scalameta is also a big change. I think it can be a major version update if this pr is merged.

brbrown25:
> Anyway, it's true that the changes are breaking changes and switching to scalameta is also a big change. I think it can be a major version update if this pr is merged.

Yeah that's what I was thinking would make sense as well. You're switch to scalameta definitely made this more manageable!

ahjohannessen:
>since this switch to scalameta is in fact a breaking change

Make it a major version change and call it a day.

brbrown25:
> > since this switch to scalameta is in fact a breaking change
> 
> Make it a major version change, add mima filters, and call it a day :)

I can add that in today!

ahjohannessen:
What is holding this from being merged, any blockers? :)

brbrown25:
> What is holding this from being merged, any blockers? :)

not that I'm aware of, I think it just needs a review.

ahjohannessen:
@mkurz WDYT? :)

mkurz:
I will review in like one week. My time currently is consumed by an old customer's project where I currently help out. I should be back to Play work by mid next week. Thanks for you patience.

brbrown25:
> I will review in like one week. My time currently is consumed by an old customer's project where I currently help out. I should be back to Play work by mid next week. Thanks for you patience.

Thanks so much! I really appreciate it.

build.sbt
```diff
@@ -112,7 +144,10 @@ lazy val plugin = project
     scalaVersion                            := Scala212,
     libraryDependencies += "org.scalatest" %%% "scalatest" % ScalaTestVersion % Test,
     Compile / resourceGenerators += generateVersionFile.taskValue,
-    scriptedLaunchOpts += version.apply { v => s"-Dproject.version=$v" }.value,
+    scriptedLaunchOpts := {
+      scriptedLaunchOpts.value ++
+        Seq("-Xmx1024M", s"-Dplugin.version=${version.value}", s"-Dproject.version=${version.value}")
```
mkurz, line 149: Is `-Dplugin.version=${version.value}` actually needed?

compiler/src/test/scala-3/play/twirl/compiler/test/Helper.scala
```diff
@@ -0,0 +1,106 @@
...
+
+    val twirlCompiler = TwirlCompiler
+
+    val classloader = new URLClassLoader(
+      Array(generatedClasses.toURI.toURL),
+      Class.forName("play.twirl.compiler.TwirlCompiler").getClassLoader,
+    )
+
+    // A list of the compile errors from the most recent compiler run
+    val compileErrors = new mutable.ListBuffer[CompilationError]
```
mkurz, line 38: Not used anymore, can be removed.

.travis.yml
```diff
@@ -28,6 +28,12 @@ jobs:
         - SCALA_VERSION=2.13.8
         - ADOPTOPENJDK=11
 
+    - name: "Run tests with Scala 3 and AdoptOpenJDK 11"
+      script: scripts/test-code.sh
+      env:
+        - TRAVIS_SCALA_VERSION=3.1.0
```
mkurz, line 34: This should be `SCALA_VERSION`, not `TRAVIS_SCALA_VERSION`

.travis.yml
```diff
@@ -40,6 +46,12 @@ jobs:
         - SCALA_VERSION=2.13.8
         - ADOPTOPENJDK=8
 
+    - name: "Run tests with Scala 3 and AdoptOpenJDK 8"
+      script: scripts/test-code.sh
+      env:
+        - TRAVIS_SCALA_VERSION=3.1.0
```
mkurz, line 52: This should be `SCALA_VERSION`, not `TRAVIS_SCALA_VERSION`

mkurz (changes requested):
Thanks for you work!
Some minor things need to be addressed.
Also I think you actually didn't run the test on Travis with the correct Scala version... So let's see how that turns out after setting the correct version.

Also, the changes in `TwirlParser` look ok to me, I know this needed change because of the `if` problem you described, however I hope this is not breaking apps.

Please update the pull request, I may review a second time and after merging, we can release 1.6.0-M2 so people can start testing.
Thanks again!

build.sbt
brbrown25: @tototoshi beating me to the feedback nice! I'll double check that one

brbrown25: Looks like it's not needed so I removed it!

mkurz:
I did some local testing:
```
$ sbt
sbt:twirl> ++3.1.1
sbt:twirl> test
...
-- [E006] Not Found Error: /home/mkurz/work/github-play-repos/twirl/compiler/target/test/twirl-parser/generated-templates/html/error.template.scala:19:63 
19 |Seq[Any](format.raw/*2.1*/("""<h1>Hello """),_display_(/*2.12*/world),format.raw/*2.17*/("""</h1>
   |                                                               ^^^^^
   |                                                        Not found: world

...
```
here is a screenshot so you see this is highlighted in red:
image

This from 
https://github.com/playframework/twirl/blob/2c064e741ff8771c68e6397518fc05cb16d61c67/compiler/src/test/scala/play/twirl/compiler/test/CompilerSpec.scala#L174-L180

However, it is not a real error, the test passes and the `CompilationError` object will correctly be thrown. I guess this is just the output of the Scala 3 compiler twirl executes... However the problem I have with that is that these log entries may confuse people... In Scala 2.12/2.13 these entries will not be displayed (because the Scala 2 compiler is used which obviously acts different).
Is it possible to suppress this log?

brbrown25:
I'm away from computer for a few days for family vacation but can look into n it when I return.

tototoshi:
I added a commit to suppress compile errors log.
@brbrown25 Have a good vacation 😄

mkurz:
@tototoshi I looked at your last commit, however you now suppress logging of compile errors only for tests? Or am I wrong? My idea was to always suppress that errors, not just for the CI tests... It would be also confusing in real apps to see that errors. WDYT?

brbrown25:
@tototoshi thank you!

tototoshi:
@mkurz
My change is only for tests. In real apps, the compilation of the scala files generated by twirl is the task of sbt-twirl.
And sbt-twirl does not support cross-build against scala3 yet (because it is sbt-plugin running only on scala 2.12).
I think the problem you are thinking about will appear in the future when sbt is running on scala 3. But as for now, I think the current state is enough.

tototoshi:
Umm... My understanding may be wrong. sbt-plugin runs on scala2.12 but the compilation of templates should be run on scala3. 
OK. I will check it later.

tototoshi:
Now I understand the problem.
It seems that the sourcePositionMappers setting provided by sbt-twirl doesn't work with Scala3. I didn't add changes to sbt-plugin, and the sources generated by twirl has no difference between Scala2 and Scala3. 
I doubt this problem comes with sbt and dotty. I will investigate them next.

tototoshi:
I found the cause.

https://github.com/lampepfl/dotty/blob/d2ebd75a2e63e3dd885e7bad94472931d0bae296/sbt-bridge/src/dotty/tools/xsbt/DelegatingReporter.java#L47

When Scala3 is compiled using sbt, compilation errors are reported by this `DelegatingReporter`. `DelegatingReporter` delegates the log output process to sbt's `ManagedLoggedReporter`. The problem seems to be that the error information passed to `ManagedLoggedReporter` includes messages generated by Dotty as `rendered` parameter of `Problem`.

If the `rendered` parameter is included in the error information, sbt will display the contents of the `rendered` parameter, and information other than the `rendered` parameter is ignored. So the error information in the Twirl template translated by `sourcePositionMappers` in `LoggedReporter` will not be displayed.

https://github.com/sbt/sbt/blob/fedf6af60d86ed39ae9c94161f13a79ffc4494e8/internal/util-interface/src/main/java/xsbti/Problem.java#L21-L28
https://github.com/sbt/zinc/blob/f6d36ff81aab6063a8ed5dd06c51ea6b9fef16f8/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/ProblemStringFormats.scala#L27-L28

If I modify the `DelegatingReporter` so that the `rendered` parameter is not passed to sbt, it is possible to display the same error message as when compiling with Scala2. However, the message stored in the `rendered` parameter is from the Dotty compiler and contains more detailed error messages, so I feel a bit uncomfortable making it completely invisible.

I think I have nothing that I can do In this Twirl repository, and ideally, either or both sbt and Dotty's sbt integration need to be improved.

tototoshi:
https://github.com/lampepfl/dotty/issues/14691

brbrown25:
@tototoshi, @mkurz  I just realized I messed up and didn't update the travis.yml so I've pushed that fix and we should get a clean travis pass on 3.1.1 now.

mkurz:
Travis not working, closeing/reopening to trigger it...

brbrown25:
@mkurz since we got green builds in Travis do you think we might be ready to merge this soon?

mkurz (approved):
OK lets merge. Like said I did some local testing, also tests on CI pass so it looks promising.  In case problems occur when can always fix them in upcoming pr/commits.
@tototoshi I saw https://github.com/sbt/zinc/issues/1074#issuecomment-1073247308, if you come up with a fix let us know.

I am too tired now, but I can release 1.6.0-M2 this week so people can play around with it, also to upgrade in other projects like Play to see if we trigger some other bugs.

mkurz:
Opened #498 to keep track of the Scala 3/sourcePositionMapper problem

mkurz:
This causes
- #711

If someone has time to look into that, I would appreciate it, thanks!

PR #477, opened 2022-01-21
CI with GitHub Actions

tototoshi:
Replaced the existing Travis configuration with the GitHub Actions configuration.
I used the play-json configuration as a reference, especially for publish.yml.
I've confirmed that the validation and testing work, but I haven't been able to confirm that the release works.

Notifications to Gitter have not been migrated. (I wonder if this Gitter channel is still active?)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,34 @@
+name: Publish
+
+on:
+  push:
+    branches: # Snapshots
+      - main
+    tags: ["*"] # Releases
+
+jobs:
+  publish-artifacts:
+    name: Publish / Artifacts
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          # we don't know what commit the last tag was it's safer to get entire repo so previousStableVersion resolves
+          fetch-depth: 0
+
+      - name: Set up JDK 8
+        uses: olafurpg/setup-scala@v13
+        with:
+          java-version: adopt@1.8
+
+      - name: Cache Coursier cache
+        uses: coursier/cache-action@v6.2
+
+      - name: Publish artifcats
+        run: sbt ci-release
+        env:
+          PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
+          PGP_SECRET: ${{ secrets.PGP_SECRET }}
+          SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
+          SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,7 +49,7 @@ jobs:
           # Avoid sbt-dynver not seeing the tag
           fetch-depth: 0
 
-      - name: Set up JDK 11
+      - name: "Set up JDK ${{ matrix.java }}"
         uses: olafurpg/setup-scala@v13
         with:
           java-version: "adopt@1.${{ matrix.java }}"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,90 @@
+name: Tests
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+
+jobs:
+  validate:
+
+    runs-on: ubuntu-20.04
+
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          # Avoid sbt-dynver not seeing the tag
+          fetch-depth: 0
+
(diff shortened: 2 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,34 @@
+name: Publish
+
+on:
+  push:
+    branches: # Snapshots
```
octonato, line 5: this will publish snapshots, which is what we want, but we will need to add 

```scala
// We are publishing snapshots to Sonatype
ThisBuild / dynverSonatypeSnapshots := true
```

To the build.sbt

.github/workflows/publish.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Cache Coursier cache
+        uses: coursier/cache-action@v6.2
+
+      - name: Publish artifcats
+        run: sbt ci-release
+        env:
+          PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
+          PGP_SECRET: ${{ secrets.PGP_SECRET }}
+          SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
+          SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
```
octonato, line 34: I will need to add the secrets. They are not there yet. 
We may want to wait a bit with that though. @mkurz is getting a new sonatype user that we will use to publish artifacts. 

Once we have the user, we can add the secrets at organisation level. So far we have done per repo, but maybe it's better to do once and for all at org level. Anyway, to follow-up...

octonato (commented):
Thanks a lot @tototoshi 

This is looking good. There are a few things missing though. See my comments below. 

Also, we will need to change mergify (see https://github.com/playframework/play-ws/pull/639/files#diff-6a30e47dd51449847422017b16878889b848f6602ac56d5a2e67ce8a19afade2). 

But maybe we should do that on a follow-up PR. So first we get it running in GH Actions and merged. Then we remove Travis and adapt Merfigy.

.github/workflows/publish.yml
mkurz: We can simply set it up once for the org, done. I'll do that as soon I have the go from sonatype staff.

.github/workflows/publish.yml
ihostage: @@octoate In fact, this is not necessary. 😉 
Because If we are using `sbt-ci-release` for publishing (for Twirl we do, as I see), it already has this property.
https://github.com/sbt/sbt-ci-release/blob/main/plugin/src/main/scala/com/geirsson/CiReleasePlugin.scala#L105
It's my mistake, that I added this property to `play-ws` explicitly 😞

tototoshi: Thanks. 
Do we only need the configuration for the main branch to publish as a SNAPSHOT version?
The travis settings include `/^\d+\.\d+(\.\d+)?(-\S*)?$/`, but I removed it because it doesn't match the branch that is currently in the twirl repository.

octonato: @ihostage, you are right. No need to add it. I forgot about it. 

@tototoshi, I think as it is in the PR is fine.

I think we have everything except the credentials to publish.

.github/workflows/publish.yml
mkurz: The secrets are set up now. Happy publishing.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          # we don't know what commit the last tag was it's safer to get entire repo so previousStableVersion resolves
+          fetch-depth: 0
+
+      - name: Set up JDK 8
+        uses: olafurpg/setup-scala@v13
+        with:
+          java-version: adopt@1.8
```
tototoshi, line 23: https://github.com/olafurpg/setup-scala/issues/49 👀  Should I use setup-java?

ihostage: I think we should try switch to [coursier/setup-action](https://github.com/coursier/setup-action)

tototoshi: or https://github.com/playframework/play-json/pull/705

brbrown25:
One suggestion I would add for this would be https://github.com/djspiewak/sbt-github-actions which can automate the yaml generation and be an extra check. But IMHO I would do that as a follow up pr

tototoshi:
sbt-github-actions is great, but it seems we have [playframework/.github](playframework/.github) now and it is used in play-json.

mkurz:
Fixed with #483
Thanks @tototoshi and @brbrown25, I opened https://github.com/playframework/play-meta/issues/219 as a reminder to look at sbt-github-actions someday...

PR #478, opened 2022-01-22
Update scalafmt to 3.3.1

tototoshi:
In recent versions of scalafmt, setting ruuner.dialect is mandatory.

The current configuration file does not have runner.dialect set.
Hence we are in a state where we cannot accept pull requests by scala-steward.
https://github.com/playframework/twirl/pull/474


mkurz:
Thanks!

PR #479, opened 2022-01-22
Add syntax related fixes for Scala3 cross build

tototoshi:
I know there are already some people working on cross-building in Scala3. But I'm going to try it too.

The following fixes have been added with this PR.
- Specify the return type for "implicit def
- Stop calling methods without "()".

We probably have a lot more work to do for the Scala3 cross-build,
I think it's better to get these fixes in first.
It will reduce noise and make it easier to work and review code.


mkurz:
:+1:

PR #485, opened 2022-01-30
Add a testcase for By-name parameters

tototoshi:
Twirl is implemented taking into account the By-Name parameter of the template argument.
https://github.com/playframework/twirl/blob/a6c8e45a797ef85847fbd94274023ff023682152/compiler/src/main/scala/play/twirl/compiler/TwirlCompiler.scala#L549

However, there is no test case for it.
I would like to add a test case to make it safer to support cross-building with Scala3.


mkurz:
Tests fail

mkurz (approved):

PR #506, opened 2022-04-19
sbt 1.7

mkurz:
This contains the Zinc fix from @tototoshi 
Let's see if that works now by removing the workaround.

### DIFF
--- a/docs/project/build.properties
+++ b/docs/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.6.2
+sbt.version=1.7.0-M2
--- a/compiler/src/test/scala-3/play/twirl/compiler/test/Helper.scala
+++ b/compiler/src/test/scala-3/play/twirl/compiler/test/Helper.scala
@@ -14,7 +14,6 @@ import Contexts.Context
 import Contexts.ctx
 import dotty.tools.dotc.Compiler
 import dotty.tools.dotc.Driver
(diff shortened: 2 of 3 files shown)


mkurz:
@tototoshi The problem still occurs: https://github.com/playframework/twirl/runs/6078515650?check_suite_focus=true#step:7:130
Or did I understand something wrong?

tototoshi:
The test does not use sbt, it directly displays the error messages passed from the Dotty compiler. 

I will create a sample project to verify the fix. Please wait a moment.

mkurz:
So the errors are still displayed: https://github.com/playframework/twirl/runs/7289710331?check_suite_focus=true#step:9:133
@tototoshi I guess this is because the problem has not been fixed in dotty? So I guess I will not remove the patch in this pull request for now...

docs/project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.6.2
+sbt.version=1.7.0
```
ihostage, line 1:
```suggestion
sbt.version=1.7.1
```
😉

mkurz:
So we don't revert the fix here. I will merge #534 instead and keep that pull request if we want to have a look later.

tototoshi:
@mkurz 
Yes, your guess is correct.
I think it is hard to solve this problem on the dotty side, so I decided to send a pull request to zinc.

PR #512, opened 2022-04-28
Handle an exception from scalameta when parsing template args fail

tototoshi:
TwirlCompiler uses scalameta to analyze template parameters.
I found that there was no error handling during the parsing by scalameta and the compiler would abort if the type was not specified in the template argument. This PR fixes it.


mkurz (approved):
LGTM Thanks!

tototoshi:
🤔

PR #519, opened 2022-05-06
Fix release by tag

mkurz:
@octonato just tagged `1.6.0-M2`, however not artifacts got pushed to maven central.

I think this is because of using double quotes...

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -10,8 +10,4 @@ jobs:
   publish-artifacts:
     name: Publish / Artifacts
     uses: playframework/.github/.github/workflows/publish.yml@v2
-    secrets:
-      username: ${{ secrets.SONATYPE_USERNAME }}
-      password: ${{ secrets.SONATYPE_PASSWORD }}
-      pgp_passphrase: ${{ secrets.PGP_PASSPHRASE }}
-      pgp_secret: ${{ secrets.PGP_SECRET }}
+    secrets: inherit
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: # Snapshots
       - main
-    tags: ['*'] # Releases
+    tags: ['**'] # Releases
 
 jobs:
   publish-artifacts:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: # Snapshots
       - main
-    tags: ["*"] # Releases
+    tags: ['*'] # Releases
 
 jobs:
   publish-artifacts:


mkurz:
See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#patterns-to-match-branches-and-tags

.github/workflows/publish.yml
```diff
@@ -4,14 +4,10 @@ on:
...
 jobs:
   publish-artifacts:
     name: Publish / Artifacts
     uses: playframework/.github/.github/workflows/publish.yml@v2
-    secrets:
-      username: ${{ secrets.SONATYPE_USERNAME }}
-      password: ${{ secrets.SONATYPE_PASSWORD }}
-      pgp_passphrase: ${{ secrets.PGP_PASSPHRASE }}
-      pgp_secret: ${{ secrets.PGP_SECRET }}
+    secrets: inherit
```
mkurz, line 13: @ihostage You probably like this new feature 😉 
* https://github.blog/changelog/2022-05-03-github-actions-simplify-using-secrets-with-reusable-workflows/
* https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow

.github/workflows/publish.yml
```diff
@@ -4,14 +4,10 @@ on:
   push:
     branches: # Snapshots
       - main
-    tags: ["*"] # Releases
```
mkurz, line 7: Not 100% sure but I think double quotes may be seen as normal string? Only when using singlel quotes a string is escaped?

.github/workflows/publish.yml
ihostage: Yeeees, cooool! 🕺

mkurz:
It seems the double quotes are not the problem.
It seems the problem is that when creating a release from the GitHub UI the push tags event will not be called...

mkurz:
Only when `git tag ..` locally and pushing the tag the workflow will start.

mkurz:
There is the `on: release` setting, however I am not sure if that is good: What if I git tag locally push (now workflow runs) and then I will push a release with that existing tag... Again a worklow will run. Not so nice.
Also what if I edit a already published release, will the published event trigger again after saving the changes? Have to test.

ihostage:
>It seems the double quotes are not the problem.
>It seems the problem is that when creating a release from the GitHub UI the push tags event will not be called...

I remember the same problem in one of Akka repo, but forgot which one exactly 🤦‍♂️ 
@octonato Maybe you remember?

ihostage:
Someone from Akka team already analysed this case, I'm sure.

mkurz:
I saw that too, there was a comment from Enno I think, I will investage.

mkurz:
I found the problem!!! 🥳 😪  I think it's a bug in GitHub... I tried a lot in https://github.com/mkurz/release-drafter-test/

Actually the `on` `push` `tags: ["*"]` setting is correct!
The problem is that when a release draft is created by release drafter, then the author of the draft is `github-actions`.
Now when publishing the release no publish event will be send - probably because that user does not have permissions to the repository. However the tag is created 🤔 but no events send...

image

I think this is a bug in GitHub, because it should not use the permissions from the author of the draft, but the permissions from the user that clicks "publish" for the release.

When I created a release draft myself (user mkurz), and then published a release via UI, it worked! the push tag event will be send and the release starts! See here for the 1.6.0-M6 tag: https://github.com/playframework/twirl/runs/6331248053?check_suite_focus=true

I will report a bug to GitHub.
Also I have an idea for a workaround already... 👍 
(Also I will clean up everything tomorrow)

PR #523, opened 2022-05-18
GitHub actions last round + badges

mkurz:


ihostage (approved):

.github/workflows/publish.yml
```diff
@@ -4,7 +4,9 @@ on:
   push:
     branches: # Snapshots
       - main
-    tags: ["*"] # Releases
+    tags: ["**"] # Releases
+  release:
+    types: [published]
```
ihostage, line 9: @mkurz Are we really need to run this workflow on `published release` event? 🤔 
It was failed because artifacts had been already published by `pushing tag` event 🤷‍♂️ 
https://github.com/playframework/twirl/actions/runs/7127596926
https://github.com/playframework/twirl/actions/runs/7127594512

mkurz: Yeah I saw that in the past. The idea back when we added that was that we can just create and publish a GitHub release, which will create the tag and then 20 minutes later or something the release is on maven central.
That was the idea.

However, we never use that approach, I think it's a bad approach, because if something goes wrong with publishing to maven central (or tests fail, compile problem, etc.) and the release never makes it to the maven central repos, people already get notified by GitHub about the release.
So that's why it's always safer to push the tag by hand, wait for maven central, and when that is ready, publish the GitHub release for the already existing tag.

I think we should remove the `on: release` event for all repos, since I consider it unsafe and bad practise.

(I know we always had/have errors but I just ignored it for now)

ihostage: > However, we never use that approach, I think it's a bad approach, because if something goes wrong with publishing to maven central (or tests fail, compile problem, etc.) and the release never makes it to the maven central repos, people already get notified by GitHub about the release.
> So that's why it's always safer to push the tag by hand, wait for maven central, and when that is ready, publish the GitHub release for the already existing tag.

I absolutely agree with you 👍

PR #564, opened 2022-11-02
use scalameta-parsers instead of scalameta

xuwei-k:
reduce dependency libraries

### Before

before

### After

after


mkurz:
Thanks!

SethTisue:
nice catch!

PR #572, opened 2022-11-15
Move `TemplateCompiler#formatImports` to `TwirlCompiler`

### LINKED ISSUE #177: Move `TemplateCompiler#formatImports` to `TwirlCompiler`
This line of code is currently SBT-specific:
https://github.com/playframework/twirl/blob/master/sbt-twirl/src/main/scala/play/twirl/sbt/TemplateCompiler.scala#L68

It would make sense for it to happen in TwirlCompiler instead so that it happens in Maven and Gradle builds as well

ihostage:
Fixes #177


mkurz (approved):
:+1:

PR #594, opened 2023-02-24
Allow defining result type for local methods (needed for Scala 3 implicits)

mkurz:
To define an implicit we need to prefix the name with `implicit`:
```scala
@implicitField = @{ ... }
```
this generates:
```scala
implicit def /*6.2*/implicitField = {{ ... }}
```

Scala 3 however complains:
```
[error] ./documentation/manual/working/javaGuide/main/forms/code/javaguide/forms/withFieldConstructor.scala.html:6:2:
[error]  result type of implicit definition needs to be given explicitly
[error] @implicitField = @{ ... }
[error]  ^
````

Until now it was not possible to define a result type, so this pull request adds support for that so we can now write:
```scala
@implicitField: FieldConstructor = @{ ... }
```
which generates:
```scala
implicit def /*6.2*/implicitField/*6.15*/:FieldConstructor = {{ ... }}
```

Works locally, tested with Play's Scala 3 pull requests, just needs some tests.


mkurz:
Added tests.

PR #602, opened 2023-04-07
fix warning

xuwei-k:
```
[warn] /home/runner/work/twirl/twirl/compiler/src/main/scala/play/twirl/compiler/TwirlCompiler.scala:525:18: method paramss in trait Def is deprecated: 4.6.0
[warn]           defdef.paramss
[warn]                  ^
[warn] one warning found
```


xuwei-k:
- https://github.com/sbt/sbt/issues/7202 😭

Issue #591, opened 2023-02-22
sbt-twirl generates code that is no longer supported and/or deprecated in Scala 3


tmiseikis:
**Scala version:** 3.2.1
**sbt-twirl version:** 1.6.0-M7

The following issues are reported by Scala 3 compiler in the files generated by sbt-twirl:
- `_` is no longer supported for a wildcard import; use `*` instead 
This construct can be rewritten automatically under -rewrite -source future-migration.

```
import _root_.play.twirl.api.TwirlFeatureImports._
import _root_.play.twirl.api.TwirlHelperImports._
```

- Unless class BaseScalaTemplate is declared 'open', its extension in a separate file should be enabled by adding the import clause 'import scala.language.adhocExtensions' or by setting the compiler option -language:adhocExtensions.
See the Scala docs for value scala.language.adhocExtensions for a discussion why the feature should be explicitly enabled.

mkurz:
@tmiseikis Would you have time to provide a PR for this? @xuwei-k @tototoshi Maybe you?

tototoshi:
To solve this problem, sbt-plugin and TwirlCompiler need to be able to select the version of Scala code that is generated from the template.

What I need to do is
- to add a configuration key for sbt-plugin like `generatedCodeScalaVersion`
- to make `TwirlCompiler.compile` generate the appropriate code for each version of Scala

ahjohannessen:
Could this not be solved by:

```
templateImports  := TwirlCompiler.defaultImports(scalaVersion.value),
```

tototoshi:
@ahjohannessen Thank you. I did exactly what you said.

PR #613, opened 2023-05-27
Added support for switching the default "import" between Scala 2 and  Scala 3

tototoshi:
Twirl has been generating code with Scala 2 syntax.
Therefore, when compiling a Scala 3 project with the "-source:future" scalac option,
the compilation would fail.

To address this issue, TwirlCompiler can now generate the appropriate code
depending on the Scala version.

The Scala version is set by build tools.
if you are using Maven or gradle, it will dynamically determine the Scala version
you are currently using and set it.
If you are using sbt, it will use the value of "scalaVersion". It can't dynamically
determine the Scala version since the version for code generation (2.12.x,
the runtime of sbt) and the version for compilation are different.

This fixes https://github.com/playframework/twirl/issues/591.


compiler/src/main/java/play/japi/twirl/compiler/TwirlCompiler.java
```diff
@@ -24,8 +24,9 @@ public class TwirlCompiler {
   public static final Set<String> DEFAULT_IMPORTS;
   static {
     Set<String> imports = new HashSet<>();
+    String scalaVersion = scala.util.Properties$.MODULE$.versionNumberString();
```
xuwei-k, line 27: I think `versionNumberString` is not useful.

```
Welcome to Scala 3.2.2 (11.0.19, Java OpenJDK 64-Bit Server VM).
Type in expressions for evaluation. Or try :help.
                                                                                                                                               
scala> scala.util.Properties.versionNumberString
val res0: String = 2.13.10
```

tototoshi: oh...I didn't know this behavior.
Is there any other better way?

ahjohannessen: @tototoshi On a second thought. I am not acquainted with usage of `TwirlCompiler.java`. If one uses Scala 3 and `TwirlCompiler.java` then I suppose there is the same issue with imports if `scala.util.Properties$.MODULE$.versionNumberString()` is not providing proper version for Scala 3.

tototoshi: @ahjohannessen 
I understand that TwirlCompiler.java is an API designed for tools like Maven and Gradle.
I think the preferred behavior is that If you use TwirlCompiler.java of Scala3 build, it will output Scala code in the Scala3 syntax, and if you use a Scala2 build, it will output Scala2 code. However, when it comes to sbt 1.x, which always runs on Scala 2.12, there is a need to output code in the Scala3 syntax from Scala 2.12 builds

tototoshi: I have no idea about a convenient method that can be a direct replacement for `versionNumberString`. However, if I place a class in src/main/scala-${version} that specifies the Scala major version and load it, I can determine whether it is Scala 2 or Scala 3."

ahjohannessen: @tototoshi Found this on the internet:

```scala
object ScalaVersion extends App {
  def props(url: java.net.URL): java.util.Properties = {
    val properties = new java.util.Properties()
    val is = url.openStream()
    try {
      properties.load(is)
      properties
    } finally is.close()    
  }

  def scala2Version: String = 
    props(getClass.getResource("/library.properties")).getProperty("version.number")
    
  def checkScala3(res: java.util.Enumeration[java.net.URL]): String = 
    if (!res.hasMoreElements) scala2Version else {
      val manifest = props(res.nextElement)
      manifest.getProperty("Specification-Title") match {
        case "scala3-library-bootstrapped" =>
          manifest.getProperty("Implementation-Version")
        case _ => checkScala3(res)
      }
    }
  val manifests = getClass.getClassLoader.getResources("META-INF/MANIFEST.MF")
    
  val scalaVersion = checkScala3(manifests)
  val javaVersion = System.getProperty("java.version")

  println(s"Scala: $scalaVersion Java: $javaVersion")
}
```

xuwei-k: >  I place a class in src/main/scala-${version} that specifies the Scala major version and load it

👍 

> Found this on the internet:

I think this does not work if classpath include `scala3-library.jar` (e.g. use sbt `CrossVersion.for2_13Use3`)

ahjohannessen: @xuwei-k I think this works:

```java
  public static final Set<String> DEFAULT_IMPORTS;
  static {
    Set<String> imports = new HashSet<>();
    String scalaVersion = play.twirl.compiler.BuildInfo$.MODULE$.scalaVersion();
    imports.addAll(JavaConverters$.MODULE$
        .seqAsJavaListConverter(play.twirl.compiler.TwirlCompiler$.MODULE$.defaultImports(scalaVersion)).asJava());
    DEFAULT_IMPORTS = Collections.unmodifiableSet(imports);
  }
```

and using sbt-buildinfo for compiler:

```scala
buildInfoKeys       := Seq[BuildInfoKey](scalaVersion),
buildInfoPackage := "play.twirl.compiler"
```

tototoshi: @ahjohannessen Thanks! I will merge it.

mkurz (approved):
Thanks LGM! Using sbt-buildinfo seems like the right approach here.
In the tests the wrong scalaVersion still was used, I fixed that (also solved merge conflicts).

PR #624, opened 2023-06-19
Enable Java Format check for Gradle Plugin

ihostage:


mkurz:
:+1: All files gradle related, so I don't look too close :wink:

PR #646, opened 2023-08-11
Sync version between Sbt and Gradle

ihostage:


mkurz:
I a not 100% sure if this is the right approach (but maybe I just miss some information and context here).
Can you explain a bit when exactly do you need the version from sbt for gradle? Is it for the release process?

The problem I see is that when the `compiler/version.properties` file does not exist, I have to run publish or publishM2 to generate it, however this triggers a snapshot release to sonatype (at least when running `publish`, I just did that). Also when e.g. checking out a git tag to reproduce things locally, I would have to run `publish` just to generate the `version.properties` files, but the `publish` command will fail because a release would already exist in the maven central repos.
Do you know what I mean?
Maybe, if you explain a bit we can figure out how to solve this (or maybe I just get things wrong).

ihostage:
Hi, Matthias! 👋 
This PR needs to us by two reason. 
1. Have the same version for Gradle plugin as other Twirl modules
2. Gradle plugin will be depends on Twirl compiler and we need a version of compiler that published locally or public.

> The problem I see is that when the compiler/version.properties file does not exist, I have to run publish or publishM2 to generate it, however this triggers a snapshot release to sonatype (at least when running publish, I just did that). 

`publishM2` installs artifacts into Local Maven repo we won't have a problem with Sonatype. Anyway, we always can create this file manually if sure that this compiler version is already published somewhere. 

> Do you know what I mean?

Yep. Think that it's just a bridge between Sbt and Gradle to tell Gradle what version Gradle plugin should be and what version compiler we should depends. Maybe in the future this will be separated variables but now I think one is enough.

build.sbt
```diff
@@ -129,7 +132,10 @@ lazy val compiler = project
     libraryDependencies += ("org.scalameta"                     %% "parsers" % "4.8.6").cross(CrossVersion.for3Use2_13),
     run / fork                                                  := true,
     buildInfoKeys                                               := Seq[BuildInfoKey](scalaVersion),
-    buildInfoPackage                                            := "play.twirl.compiler"
+    buildInfoPackage                                            := "play.twirl.compiler",
+    publishM2                                                   := publishM2.dependsOn(saveCompilerVersion).value,
+    publish                                                     := publish.dependsOn(saveCompilerVersion).value,
+    publishLocal                                                := publishLocal.dependsOn(saveCompilerVersion).value
```
mkurz, line 138: +1 I was thinking about `publishLocal` as well.

mkurz (approved):
Should be ok, at least does not destroy anything :wink:

PR #662, opened 2023-08-30
[Gradle] Add template imports to the action params

ihostage:


gradle-twirl/src/main/java/play/twirl/gradle/TwirlPlugin.java
```diff
@@ -44,24 +47,33 @@ public TwirlPlugin(ObjectFactory objectFactory) {
...
+    project.afterEvaluate(
+        __ -> {
+          Dependency twirlCompiler =
+              project
+                  .getDependencies()
+                  .create(
+                      String.format(
+                          "com.typesafe.play:twirl-compiler_%s:%s",
+                          twirlExtension.getScalaVersion().get(), twirlVersion));
+          conf.defaultDependencies(dependencies -> dependencies.add(twirlCompiler));
```
vlsi, line 75: have you tried moving `dependencies.create(...)` into `defaultDependencies` lambda?
Then `afterEvaluate` will not be required.

PR #676, opened 2023-09-12
Change groupId to `org.playframework.twirl`

ihostage:
Before add Gradle Plugin we should finalize a groupId that also be used in Gradle Plugin Portal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ To add the sbt plugin to your project add the sbt plugin dependency in
 `project/plugins.sbt`:
 
 ```scala
-// twirl 1.7:
+// twirl 2.0 and newer:
 addSbtPlugin("org.playframework.twirl" % "sbt-twirl" % "LATEST_VERSION")
 // twirl 1.6:
 addSbtPlugin("com.typesafe.play" % "sbt-twirl" % "1.6.1")
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 [![OpenCollective](https://img.shields.io/opencollective/all/playframework?label=financial%20contributors&logo=open-collective)](https://opencollective.com/playframework)
 
 [![Build Status](https://github.com/playframework/twirl/actions/workflows/build-test.yml/badge.svg)](https://github.com/playframework/twirl/actions/workflows/build-test.yml)
-[![Maven](https://img.shields.io/maven-central/v/com.typesafe.play/twirl-api_2.13.svg?logo=apache-maven)](https://mvnrepository.com/artifact/com.typesafe.play/twirl-api_2.13)
+[![Maven](https://img.shields.io/maven-central/v/org.playframework.twirl/twirl-api_2.13.svg?logo=apache-maven)](https://mvnrepository.com/artifact/org.playframework.twirl/twirl-api_2.13)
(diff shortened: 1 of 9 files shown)


ihostage:
@mkurz @gmethvin WDYT?

gmethvin:
Any reason not to just use `org.playframework` for everything?

ihostage:
I think we shouldn't use top-level group `org.playframework` for independent project like Anorm and Twirl. 
Anorm, for example, already use group `org.playframework.anorm` ([source link](https://github.com/playframework/anorm/blob/35483c8ad3c6946b79d3b4411a5c06bf6a2514be/project/Common.scala#L23), [repo link](https://repo1.maven.org/maven2/org/playframework/anorm/)) and I encourage that and think we should use the same approach for Twirl. `org.playframework` will be use for Play itself.

mkurz:
Actually I was thinking to keep `com.typesafe` for the 1.6 release (for which we published RC's already) and switch to ´org.playframework.[twirl]` afterwards.

ihostage:
Yes, @mkurz, I also thought about this case. And that's all right for my. Just I thought if we haven't published final release 1.6 yet we can do rename by one step instead of two and save our time 😄

mkurz:
> I think we shouldn't use top-level group `org.playframework` for independent project like Anorm and Twirl. 

The question here is what is an independent project? play-json also is an independent project, it does not depend on any of the Play (core) projects. play-ws also, it just depends on play-json. cachecontrol the same, no dependencies.
So following this rule we would need `"org.playframework.play-json" %% "play-json"`, `"org.playframework.cachecontrol" %% "cachecontrol"` etc.

> Anorm, for example, already use group `org.playframework.anorm`

I know, anorm always was a bit special... Also they were the only one for a long time to publish in the `org.playframework` groupId` already...

> Group `org.playframework` will be use for Play itself.

Again the question what is Play itself? I already published play-ws 3 m1 (https://repo1.maven.org/maven2/org/playframework/play-ws_3/) under the "normal" groupId.

I understand for twirl it would make some sense, but I think it's more because of the name "twirl"... People use "play-json" without play also, same with "play-ws", people use it wilthout a Play application....

ihostage:
Yes, Matthias, I understand what you mean. Personally, I feel inwardly "independent" by next property:
* Don't have a `play` suffix in the name
* Is multi-module project, otherwise we don't need special groupId for one artifact.

And in fact except Anorm and Twirl I don't have other examples until 🤷‍♂️

mkurz:
OK, we can do that. So if artifacts start with `play-` (because they are play project related) we keep in top-level. If not starting with `play-` AND multi-module than in sub-group (like anorm, twirl, actually also netty-reactive-streams, see https://github.com/playframework/netty-reactive-streams/pull/199). If not starting with `play-` AND _NOT_ multi-module than also top-level (like `cachecontrol`. Actually would could rename it to `play-cachecontrol` but that does not make sense because it is totally independent of everything and a standalone project, it only depends on `slf4j`, also others use that library AFAIK, independent of Play).
In the end it's "just" naming, not a feature or bug fix...

ihostage:
It's ok to me. I agree 🤝

PR #678, opened 2023-09-12
Gradle Support

### LINKED ISSUE #83: Gradle Support
Currently Twirl only works on sbt, is there any work to also have a working gradle/maven plugin?

ihostage:
Fixes #83


ihostage:
Depends on #676

ihostage:
@mkurz Gradle Plugin can also be a part of final 2.0.0 😉

mkurz:
I did not review yet, but why is that needed? What about https://gradle.github.io/playframework (https://github.com/gradle/playframework)? Can this not be used?

ihostage:
Unfortunately this plugin more dead than alive 😞 Of course, I inspired them while develop, but I want that Play have a first-class support not only Sbt but also Gradle and Maven. And support for Twirl is a first step in this way.

mkurz:
Hm.. I know it took a while until they supported Play 2.8 (I was following the repo a bit, see https://github.com/gradle/playframework/pull/163)
Also I was helping them when something did not work:
* https://github.com/gradle/playframework/pull/183#issuecomment-1490497661
* https://github.com/gradle/playframework/issues/109#issuecomment-1483105358

So for me it seems they did some work. But anyway, I don't know in which state the project is in general, if something is missing.

Just one more question: If we merge this PR, does the current Play Gradle plugin (https://gradle.github.io/playframework/) still work? Like, what I mean, are the compatible, like can they life in parallel? Or will your code here make the "Gradle Play Plugin" unusable?

ihostage:
> Hm.. I know it took a while until they supported Play 2.8 (I was following the repo a bit, see https://github.com/gradle/playframework/pull/163)

Yes, I also watch this repo and I'm upset a current status 😞 I don't believe that something change in Gradle team strategy related with this plugin. For example Gradle Play plugin is very unfriendly for Gradle Kotlin DSL.

> Just one more question: If we merge this PR, does the current Play Gradle plugin (https://gradle.github.io/playframework/) still work? Like, what I mean, are the compatible, like can they live in parallel? Or will your code here make the "Gradle Play Plugin" unusable?

I didn't test a both plugin in one project 🤷‍♂️ I don't think that someone want to use it both in one project right now. I think when I'll finish a work on "official" Play Gradle Plugin users can migrate from Gradle Play plugin => Play Gradle Plugin + Twirl Gradle plugin. 🤞

mkurz:
OK, I will take a look at this on the weekend and promise if it's ok to merge before 2.0.0 final.

ihostage:
Thanks, Matthias! 
PS: Moreover, I think instead of support Twirl 2.0.0 in Gradle Play plugin (I mean changing `groupId` in https://github.com/gradle/playframework/blob/818d998bcca6bce1d2e8d7deaaecaf7a19be55a1/src/main/java/org/gradle/playframework/tools/internal/twirl/TwirlCompilerAdapterV10X.java#L79) they should just drop all Twirl source part from plugin and recommend to use our plugin. :blush: As a bonus, our plugin supports [Gradle Build Cache](https://docs.gradle.org/current/userguide/build_cache.html) and [Incremental build](https://docs.gradle.org/current/userguide/incremental_build.html) 😄

mkurz (approved):
LGTM!
I added `GRADLE_PUBLISH_KEY` and `GRADLE_PUBLISH_SECRET` as organization secrets, let's see if it works.

mkurz:
@ihostage I published 2.0.0-M3, looks good so far, but we have to wait for approval:
https://github.com/playframework/twirl/actions/runs/6602032175/job/17933650062#step:6:167
```
Publishing plugin org.playframework.twirl version 2.0.0-M3
Thank you. Your new plugin org.playframework.twirl has been submitted for approval by Gradle engineers. The request should be processed within the next few days, at which point you will be contacted via email.
```

ihostage:
Awesome 🥳 
Thank you, Matthias! 👍

PR #694, opened 2023-10-06
Adoption to new `groupId` in Twirl sbt plugin

ihostage:


mkurz:
Better than mine (#693) :+1:

PR #709, opened 2023-11-02
Set `mimaPreviousArtifacts` to last "major" release

mkurz:


build.sbt
```diff
@@ -16,11 +16,10 @@ def parserCombinators(scalaVersion: String) = "org.scala-lang.modules" %% "scala
   }
 }
 
+val previousVersion: Option[String] = Some("2.0.1")
```
mkurz, line 19: 2.0.0 was not published: https://github.com/playframework/twirl/releases/tag/2.0.0
(because sbt-dynver read the wrong tag on a commit that had two tags...)

Issue #711, opened 2023-11-13
A whitespace after if(){}else{} is not showing


yousuketto:
A whitespace after `if(condition) {...} else {...}` is not showing.

## Reproduction
scalaVersion = "2.13.12"
sbt-twirl version: `1.6.2` (Reproduced in `1.6.0-M1+38-e141ce17-SNAPSHOT`, Not reproduced in `"1.6.0-M1+35-de50d09a-SNAPSHOT`)

### code
```
$ cat src/main/twirl/com/github/yousuketto/sample.scala.txt
@(b: Boolean)
You got @if(b) {a pullrequest} else {pullrequests} from someone.
```

### Execute in sbt console
```
scala> com.github.yousuketto.txt.sample(false)
val res0: play.twirl.api.TxtFormat.Appendable =
You got pullrequestsfrom someone.

scala> com.github.yousuketto.txt.sample(true)
val res1: play.twirl.api.TxtFormat.Appendable =
You got a pullrequestfrom someone.

scala>
```

### Expected
```
scala> com.github.yousuketto.txt.sample(true)
val res0: play.twirl.api.TxtFormat.Appendable =
You got a pullrequest from someone.

scala> com.github.yousuketto.txt.sample(true)
val res1: play.twirl.api.TxtFormat.Appendable =
You got a pullrequest from someone.

scala>
```

mkurz:
According to
```
git log de50d09a..e141ce17
```

this must be caused by e8b6ceb2542f90e7c1baa18cea5fcc3688a0d3ad

@tototoshi or also @yousuketto can you take a look here and provide a pull request if you have time? Also a test would be great, thanks! I may fix that, but have lots of other stuff to do, so it might take a while.

mkurz:
Fixed by
- #712

PR #712, opened 2023-11-14
Fix whitespaces after `else{...}` and `elseif{...}` is not shown

yousuketto:
The parser does not recognize whitespace after `else {...}` as a plain, so it is not showing.

Refs #711

### DIFF
--- a/parser/src/main/scala/play/twirl/parser/TwirlParser.scala
+++ b/parser/src/main/scala/play/twirl/parser/TwirlParser.scala
@@ -73,7 +73,7 @@ import scala.util.parsing.input.OffsetPosition
  *   safeExpression : '@' parentheses
  *   ifExpression : '@' "if" parentheses expressionPart (elseIfCall)* elseCall?
  *   elseCall : whitespaceNoBreak? "else" whitespaceNoBreak? expressionPart
- *   elseIfCall : whitespaceNoBreak? "else if" parentheses expressionPart whitespaceNoBreak?
+ *   elseIfCall : whitespaceNoBreak? "else if" parentheses whitespaceNoBreak? expressionPart
  *   chainedMethods : ('.' methodCall)+
  *   expressionPart : chainedMethods | block | (whitespaceNoBreak scalaBlockChained) | parentheses
  *   expression : '@' methodCall expressionPart*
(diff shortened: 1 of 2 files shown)


ihostage (approved):

mkurz (changes requested):
LGTM, however I think the last `whitespaceNoBreak?` has to be removed now from
https://github.com/playframework/twirl/blob/6ee921152c206db21d1b8ba7fc7edca8b686ebec/parser/src/main/scala/play/twirl/parser/TwirlParser.scala#L75
or not?

Or actually... isn't following correct actually?
```
  *   elseCall : whitespaceNoBreak? "else" whitespaceNoBreak?  expressionPart
```

yousuketto:
Oh, I didn't notice the document. As you say, it has to be changed as the following. Thanks.
```
  *   elseCall : whitespaceNoBreak? "else" whitespaceNoBreak? expressionPart
```

yousuketto:
Furthermore, I noticed that elseIfCall has the same bug as elseCall. I'll fix it too.
```
@(b: Boolean, b2: Boolean)
x@if(b) {ifbody} <-whitespace
x@if(b) {ifbody} else if(b2) {elseifbody} <-whitespace
```
```
scala> com.github.yousuketto.txt.sample(true, false)
val res0: play.twirl.api.TxtFormat.Appendable =
xifbody <-whitespace
xifbody<-whitespace

scala> com.github.yousuketto.txt.sample(false, false)
val res1: play.twirl.api.TxtFormat.Appendable =
x <-whitespace
x<-whitespace

scala> com.github.yousuketto.txt.sample(false, true)
val res2: play.twirl.api.TxtFormat.Appendable =
x <-whitespace
xelseifbody<-whitespace
```

yousuketto:
I also fixed a bug in elseIfCall. If we need to split this pull request, I'll split it. Please let me know.

mkurz:
Perfect, thanks!

mkurz:
@Mergifyio backport main

mkurz:
Thanks! This is now included in just released twirl 1.6.3 and 2.0.2
- https://github.com/playframework/twirl/releases/tag/1.6.3
- https://github.com/playframework/twirl/releases/tag/2.0.2

yousuketto:
It's so helpful. Thanks for releasing.

PR #719, opened 2023-12-07
Fix format imports for Gradle/Maven

ihostage:
OMG, it's so implicitly 🤪 
Implemented here, but used in https://github.com/playframework/playframework/blob/2af8d72bb1f693dc8d7c0f66f31c20f94bcfeae6/dev-mode/play-build-link/src/main/java/play/TemplateImports.java#L24


mkurz:
ah I see more than that

mkurz (approved):
I only care about `compiler/src/main/java/play/japi/twirl/compiler/TwirlCompiler.java` which is LGTM :wink:

mkurz:
Will release as 2.0.3, let me check if we need to upgrade other dependencies...

PR #737, opened 2024-02-22
compiler: account for Scala 2 and 3 differences in generated varargs code

ahjohannessen:
This adds a bit of redudant overloads. The most important aspect for me now is that mima and tests pass plus a minor release can be shipped. At a later time, one can remove overloads and duplication, do a cleanup with a breaking change and bump major version or similar.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,14 @@ project/plugins/project/
 .bloop/
 
 compiler/version.properties
+
+.vscode/
+
+# Metals
+.metals/
+.bloop/
+project/metals.sbt
+metals.sbt
+
+# BSP
+.bsp/*
--- a/.gitignore
+++ b/.gitignore
@@ -14,8 +14,6 @@ project/plugins/project/
 .scala_dependencies
 .idea
 
-.bloop/
-
 compiler/version.properties
 
 .vscode/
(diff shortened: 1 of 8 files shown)


ihostage:
@ahjohannessen @mkurz I checked this for Gradle and as I expected all is fine. The changes in twirl-compiler works fine for all supported build systems.

ahjohannessen:
@ihostage That is great to hear :)

mkurz (approved):
Thanks!

I pushed a commit which adds an overloaded function so we stay compatible with previous releases. I know the plugin project is not checked by MiMa because we ususally don't guarantee compatibility for sbt plugins, but this is a cheap thing to do and avoids breaking apps in case someone relies on the method.
I just used a dummy scala version that makes sure Scala 2 code gets generated by default, just like before.

I will cut a new release after that.

mkurz:
@Mergifyio backport 1.6.x

ahjohannessen:
@mkurz Super :) Next thing could perhaps be to introduce something like:

```scala
trait TwirlCompiler {
   def compile(…)
   // perhaps not necessary
   def compileVirtual(…)
}
```

and do some cleanup with compile options.

mkurz:
@ahjohannessen Pull requests always welcome :wink:

Issue #745, opened 2024-03-06
Gradle Plugin: Documentation of a few gotchas


Nezisi:
I migrated a few of our repositories from SBT to Gradle thanks to the new shiny plugins ;) 

Our code base in SBT looked like this:

`<project name>/app/<java package directories>/templates/`
(for example: "example-project/app/org/bla/foobar/client/templates/")

It was based on the Play layout plugin and had the twirl templates mixed up in the "app" source set. All files, except the Twirl Templates, are written in Java.

As we wanted to migrate to a Gradle only approach and get rid of the Play layout (as it is a major source of confusion), I made a tiny mistake that cost me a few hours: 

As *all* files except the Twirl templates were in Java, I converted the source project to a Java sourceset:

`<project name>/src/main/java/<java package directories>/templates/`
(for example: "example-project/src/main/java/org/bla/foobar/client/templates/")

Which didn't work out at all... The plugin didn't even start. 
The solution to the puzzle was to convert the project to a proper scala project with the scala plugin and the scala sourceset.

It worked like a charm ;) Would be I think a nice warning for those migrating Java Projects, as it can save a lot of frustration.

Or am I missing something?

The second gotcha was that - despite the twirl templates being small - the Gradle daemon process loved too OOM. 

A link to https://docs.gradle.org/current/userguide/scala_plugin.html#sec:scala_compiling_in_external_process and maybe https://docs.gradle.org/current/userguide/build_environment.html ("org.gradle.jvmargs=(JVM arguments)") would be nice.

(for the play framework gradle plugin, too)

ihostage:
> I migrated a few of our repositories from SBT to Gradle thanks to the new shiny plugins ;)

Hi, @Nezisi! 👋 
Firstly, thank you a lot that try to use our new Gradle plugin!

> As _all_ files except the Twirl templates were in Java, I converted the source project to a Java sourceset:
> 
> `<project name>/src/main/java/<java package directories>/templates/` (for example: "example-project/src/main/java/org/bla/foobar/client/templates/")
> 
> Which didn't work out at all... The plugin didn't even start.

Can you create a small project with state that reproduce your problem before you fixed it? 🤔 

> The second gotcha was that - despite the twirl templates being small - the Gradle daemon process loved too OOM.

Currently, I think that a OOM problems can be related with https://github.com/playframework/playframework/pull/10939. When we merge this PR and if problem doesn't go out, we will spend a time for detailed analyzing a root cause of new OOM problem.

[...]

PR #746, opened 2024-03-06
[Gradle] Use `processIsolation` instead of `classLoaderIsolation`

ru5j4r0:
Hi.
First of all, I appreciate your great work.
This is my first PR for the Play projects, so sorry in advance if I'm doing something wrong.

I was trying to migrate a few Play Framework projects to Gradle, but I couldn't migrate some of them because `compileTwirl` task gave me an out of memory error of JVM's Metaspace.
I was able to migrate a project that has dozens of views, but I couldn't which has around 1,000 views.
My PC has 32GB RAM, but increasing Xmx nor MaxMetaspaceSize didn't fix the issue.

After searching for a bit, I found [this Gradle's issue](https://github.com/gradle/gradle/issues/18313).
I forked the Twirl repo, replaced `classLoaderIsolation` with `processIsolation`, published it to `mavenLocal()` and using that SNAPSHOT version of the Twirl plugin from my local Maven repo actually fixed the problem.

I just filed a PR right away because it's a small change.
What are your thoughts on this change?

**Possibly related**
- https://github.com/playframework/twirl/issues/745
  - Specifically about OOM part

**Other projects doing the same fix**
- [sigstore-java](https://github.com/sigstore/sigstore-java/pull/283)
- [ktfmt-gradle](https://github.com/cortinico/ktfmt-gradle/pull/206)

### DIFF
--- a/gradle-twirl/src/main/java/play/twirl/gradle/TwirlCompile.java
+++ b/gradle-twirl/src/main/java/play/twirl/gradle/TwirlCompile.java
@@ -72,7 +72,7 @@ public abstract class TwirlCompile extends DefaultTask {
       if (change.getFileType() == FileType.DIRECTORY) continue;
       WorkQueue workQueue =
           getWorkerExecutor()
-              .classLoaderIsolation(spec -> spec.getClasspath().from(getTwirlClasspath()));
+              .processIsolation(spec -> spec.getClasspath().from(getTwirlClasspath()));
 
       Map<String, String> templateFormats = getTemplateFormats().get();
       RelativeFile sourceFile =


ru5j4r0:
I suppose the drawback of using `processIsolation` is a increased overhead due to the need to start a new process?
Perhaps I can add an option to choose the isolation method, in case replacing `classLoaderIsolation` with `processIsolation` by default is not acceptable.

ihostage:
Hi, @ru5j4r0! 
> I suppose the drawback of using processIsolation is a increased overhead due to the need to start a new process?

It's okay, don't worry. I will investigate this issue in more detail next week.

ihostage (approved):
Unfortunately, I couldn't find a way to write a unit test for this case 😞 
But I've made a lot of experiments locally and I agree with the proposed changes.
Thank you very much, @ru5j4r0, and apologies for the delay!

ru5j4r0:
@ihostage 
I’m the one who owes you an apology for opening a PR without tests or reproducible examples. Thank you for all your work!

Issue #722, opened 2023-12-18
Support for more than 22 arguments in the twirl template


JamesWilson3:
Hi @mkurz,

I read that Scala 3 has now removed the restrictions regarding the 22 parameter limits (https://docs.scala-lang.org/scala3/reference/dropped-features/limit22.html)

I have also upgraded to using Play 3 with the Scala 3, but I am still encountering the error (when trying to have 23 arguments in the template):

`play.sbt.PlayExceptions$CompilationException: Compilation error[type Template23 is not a member of play.twirl.api`

Just wondering whether the removal of the 22 parameters limit in Scala 3 has any bearing on the 22 parameter limitations in the template, and if there are any additional changes that needs to be done to make the template support for greater than 22 parameters.

mkurz:
Might be possible that twirl tries to detect how many args are given and based on this is generates the scala files, not sure, just looking at https://github.com/playframework/twirl/blob/main/api/shared/src/main/scala/play/twirl/api/Template.scala
So it looks like twirl needs to be fixed to support Scala 3 removal of the 22 parameter limit?
Would you want to work on this and provide a PR?

mkurz:
If you use Play Java, then just stay on Scala 2.13 for now. If you don't have any Scala code in your project it will make absolutely no difference for you which Scala version you use.
And then just watch this issue until it gets resolved 😉

mkurz:
So I couldn't resist and came up with 
- #747

as experiment to see if that works.

mkurz:
- Fixed in #747
- Will be part of Twirl's next patch releases 2.0.5 and 1.6.6 and Play's 3.0.3/2.9.3

PR #747, opened 2024-03-08
Support >22 args in Scala 3

mkurz:
~~Try to fix~~ Fixes
- #722

~~This is WIP.~~

If this works
- [ ] ~~the`Template.scala` should be moved into a `scala-2.13` (also `scala-2.12`?) folder and~~
- [x] in the twirl compiler we can just make use of the `ScalaCompat` `sc` parameter to decide wheter we need to add `... with _root_.play.twirl.api.TemplateX[...]` or not to stay compatible with Scala 2.13.
- [ ] ~~Also MiMa filters need to be adjusted.~~


mkurz:
Fix was quite easy and we can even ship it in a patch release.
Also added tests :+1:

mkurz:
@Mergifyio backport 1.6.x

PR #758, opened 2024-04-25
Pin Scala LTS version for Steward

ihostage:


ihostage:
@mkurz Do you know, why Steward haven't created update to 3.4 yet? 🤔

mkurz:
> @mkurz Do you know, why Steward haven't created update to 3.4 yet? 🤔

hmm.. maybe because in #751 it upgraded to from 3.3.1 to 3.3.3 and maybe on next visit it would then upgrade to 3.4.
I think scala-steward first tries to upgrade to latest patch version, then it upgrades to latest minor version and so on. But not 100% sure.

ihostage:
I have the same assumption. Anyway this pin will be useful for us 😄

ihostage:
@Mergifyio backport 1.6.x

mkurz:
> @mkurz Do you know, why Steward haven't created update to 3.4 yet? 🤔

Here is the solution: https://github.com/scala-steward-org/scala-steward/blob/ddfdff8d6a9f587f4bc452a6cf10cb76734fbaae/modules/core/src/main/resources/default.scala-steward.conf#L33-L36

```
updates.ignore = [
   ...

  // No upgrades to non LTS versions
  { groupId = "org.scala-lang", artifactId = "scala3-compiler",     version = { prefix = "3.4." } },
  { groupId = "org.scala-lang", artifactId = "scala3-library",      version = { prefix = "3.4." } },
  { groupId = "org.scala-lang", artifactId = "scala3-library_sjs1", version = { prefix = "3.4." } },

  ...
```

ihostage:
Cooool! 😄 Don't need to pin LTS in all repos 🥳

mkurz:
We can keep this two pull requests because one day the main branch will become 2.0.x and there we will not want scala to be upgraded anymore. But yes we do not need to add the pins everywhere.

PR #904, opened 2025-02-23
[2.0.x] Scalameta parsers 4.13.2 - no more `for3Use2_13` workaround

mkurz:
- Backport of #866

I looked into the scalameta parsers projects - it should be ok to just upgrade to the latest version as long as our tests pass.

Also good thing is that we can now also use the native Scala 3 artifact and starting with 4.9.9 (which we ship in latest twirl) parsers does not depend on `protobuf-java` anyomore which has vulnerabilities.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -116,13 +116,13 @@ lazy val compiler = project
       }
     },
     libraryDependencies += parserCombinators(scalaVersion.value),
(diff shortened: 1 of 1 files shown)


mkurz:
@Mergifyio backport 1.6.x

Issue #894, opened 2025-02-17
`Implicit parameters should be provided with a using clause` warning since Scala 3.7


xuwei-k:
### `build.sbt`

```scala
enablePlugins(SbtTwirl)

scalaVersion := "3.7.0-RC1-bin-20250215-43f8cdb-NIGHTLY"
```

### `project/build.properties`

```properties
sbt.version=1.10.7
```

### `project/plugins.sbt`

```scala
addSbtPlugin("org.playframework.twirl" % "sbt-twirl" % "2.1.0-M2")
```

### `src/main/twirl/example.scala.html`

```
@(x: Int)(implicit y: Int)

<p>hello</p>
```

## problem

`sbt compile`

```
/src/main/twirl/example.scala.html:3:13: Implicit parameters should be provided with a `using` clause.
[warn] This code can be rewritten automatically under -rewrite -source 3.7-migration.
[warn] To disable the warning, please use the following option: 
[warn]   "-Wconf:msg=Implicit parameters should be provided with a `using` clause:s"
[warn] <p>hello</p>
[warn]             ^
[warn] one warning found
```

## generated code

```scala
  /**/
  def apply/*1.2*/(x: Int)(implicit y: Int):play.twirl.api.HtmlFormat.Appendable = {
    _display_ {
      {


Seq[Any](format.raw/*2.1*/("""
"""),format.raw/*3.1*/("""<p>hello</p>
"""))
      }
    }
  }

  def render(x:Int,y:Int): play.twirl.api.HtmlFormat.Appendable = apply(x)(y)

  def f:((Int) => (Int) => play.twirl.api.HtmlFormat.Appendable) = (x) => (y) => apply(x)(y)

  def ref: this.type = this
```

## expect

```diff
<   def render(x:Int,y:Int): play.twirl.api.HtmlFormat.Appendable = apply(x)(y)
---
>   def render(x:Int,y:Int): play.twirl.api.HtmlFormat.Appendable = apply(x)(using y)
29c29
<   def f:((Int) => (Int) => play.twirl.api.HtmlFormat.Appendable) = (x) => (y) => apply(x)(y)
---
>   def f:((Int) => (Int) => play.twirl.api.HtmlFormat.Appendable) = (x) => (y) => apply(x)(using y)
```

## note

- https://github.com/scala/scala3/pull/22441

mkurz:
Thanks for the report. We probably want to adjust the generated scala 3 code. PRs welcome 😉

[...]

PR #933, opened 2025-05-11
Prepend "using" at the callsite when definition site has implicit parameter 

ajafri2001:
Fixes [issue-894](https://github.com/playframework/twirl/issues/894), [issue-839](https://github.com/playframework/twirl/issues/839)


ajafri2001:
I should add tests and only prepend "using" for scala3+ :sweat_smile:.  

I'm not sure if this pr is the correct way to solve the issue tbh, if you have something particular in mind I'm willing to work on it.

compiler/src/main/scala/play/twirl/compiler/TwirlCompiler.scala
```diff
@@ -162,8 +162,13 @@ case class GeneratedSourceVirtual(path: String) extends AbstractGeneratedSource
 
 object TwirlCompiler {
 
+  var isScala3 = false
+
```
tototoshi, line 166: I think this `var isScala3` is not necessary.
If you declare `emitScala3Sources` as a val, it can be referred to later as sc.emitScala3Sources.

compiler/src/main/scala/play/twirl/compiler/TwirlCompiler.scala
```diff
@@ -732,7 +733,30 @@ package """ :+ packageName :+ """
               }
               .mkString(",") + ")"
           )
+          .zipWithIndex
+          .map { case (groupStr, idx) =>
+            // Finds last element and prepends "using" if the definition site has implicit parameter eg - apply(x)(y) becomes apply(x)(using y)
+            if (TwirlCompiler.isScala3 && hasContextParameters && idx == params.size - 1)
+              groupStr.replace("(", "(using ")
+            else groupStr
+          }
```
tototoshi, line 742: This seems to be working well.
However, I thought it might be more readable if the processing were completed within the previous map { ... } block, rather than using the `zipWithIndex.map`.

tototoshi:
@ajafri2001 
Hi. I tried out this pull request, and as far as I can tell, it seems to be working well.
I left a few minor comments—please take a look.
Also, I think it would be good to add a test case to CompilerSpec.
You can find various test files in compiler/src/test/resources/, so I recommend referring to those as examples.

ajafri2001:
@tototoshi Hi, I'm unable to get the CompilerSpec Tests to compile with scala3 instead of scala2, I have added the tests but they now fail.

compiler/src/main/scala/play/twirl/compiler/TwirlCompiler.scala
tototoshi: I thought it wouldn’t be necessary to check whether a parameter is the last, since an implicit parameter is always the last parameter group.
However, I think your current implementation is fine. thanks.

ajafri2001:
@tototoshi 

>the TwirlParser probably only looks at parentheses and doesn’t check for implicit or using keywords. So I don’t think any changes are needed there.
> As for the TwirlCompiler, I think your current fix could be extended slightly to support this case too.

Yes, you were right this worked 🙇‍♂️, twirl is now able to compile the syntax `@(....)(using some_name: some_type)`.

The test fails now because `sbt test` by default runs with scala2, and I'm not sure how to test with scala3

tototoshi:
You can do it like this:

```
sbt
> ++3.3.5
> test
```

or

```
sbt ++3.3.5 test
```

tototoshi:
If you want a test to run only on Scala 3, you can place it under src/test/scala-3.
However, that approach requires splitting tests by class, which might be a bit cumbersome.

Alternatively, a simpler way is to check the Scala version at runtime and decide whether to run the test.
With ScalaTest, you can use `assume` like this:

`assume(BuildInfo.scalaVersion.startsWith("3."), "This test is only for Scala 3")`

ajafri2001:
Thanks the tests pass now

mkurz:
@Mergifyio backport 2.0.x 1.6.x

PR #947, opened 2025-06-15
add `-Xsource:3` option

xuwei-k:
https://docs.scala-lang.org/scala3/guides/migration/tooling-scala2-xsource3.html


project/Common.scala
```diff
@@ -38,12 +38,17 @@ object Common extends AutoPlugin {
     CrossVersion.partialVersion(version) match {
       case Some((2, n)) if n < 12 =>
         scalacParameters ++ Seq(
+          "-Xsource:3",
```
ihostage, line 41: @mkurz Does it not a change in dead code? 🤔

PR #981, opened 2025-07-08
Add support for ScalaNative in api

sake92:


Quafadas:
I was curious about whether twirl supported native and came across this... will keep any eye on it.

PR #1127, opened 2026-02-28
revert scala version for sbt 2

xuwei-k:
- https://github.com/scala/scala3/issues/25406


mkurz:
Ahhh, great, this maybe why my Play experiments with sbt 2 failed locally - I got the same error when running a sample Play locally with sbt2... So I guess I will downgrade all sbt plugins to 3.8.1 for now.

xuwei-k:
- Maybe the twirl-compiler needs Scala 3.8.2 (tests fail if Scala 3.8.1)
- The twirl sbt plugin depends on the twirl-compiler

😢

PR #1202, opened 2026-06-15
[2.0.x] Stabilize Gradle Twirl import order

### LINKED ISSUE #1196: Genarated templates have unstable import order and hash
Testing the plugin with gradle 8.14 and twirl version 2.0.9

We noticed back to back project reruns, with build cache enabled, triggered unexpected recompilations.

After some digging, it seems that the twirl templates generated using the `compileTwirl` task has a fingerprint that is changing every time it is invoked. This is caused by:
* the order of the default imports which is not deterministic
* the comment containing the HASH, which is changing because of the above

We worked around the issue with some changes in the `doLast` block of the `compileTwirl` task:
```scala
compileTwirl {
        doLast {
            logger.debug("Removing non-deterministic imports and comments from generated Twirl templates...")
            outputs.files.asFileTree.each { file ->
                logger.debug("Processing generated Twirl template: ${file.path}")

                def fileText = file.text
                // Identify the block of import statements
                def importBlockPattern = /(?m)(^import .*\n?)+/
                // Replace the block with the same imports sorted alphabetically
                def matcher = fileText =~ importBlockPattern

                if (matcher.find()) {
                    def fullBlock = matcher.group(0) // This is guaranteed to be a single String

                    // Split, sort, and reassemble
                    def sortedBlock = fullBlock.tokenize('\n')
                            .collect { it.trim() }
                            .findAll { !it.isEmpty() }
                            .unique()
                            .sort()
                            .join('\n') + '\n'

                    // Replace the old block with the sorted block and save
                    file.text = fileText.replace(fullBlock, sortedBlock)
                    logger.debug("Imports sorted successfully!")
                } else {
                    logger.error("No import block found.")
                }
                // Remove the non-deterministic comment that cause spurious recompilations and write back to the file
                file.text = file.text.replaceAll("(?m)^\\s{18}(HASH): .*\n", "")
                logger.debug("Non-deterministic comment removed successfully!")
            }
        }
    }
```

Would it be possible to sort the default imports before creating the template? That should automatically fix the hash calculation

mkurz:
- Fixes #1196

Gradle receives template imports through a `SetProperty`, so the order can vary between runs. That order is part of both the generated import block and the template hash.

Sort the Gradle import list before formatting it, and keep the Java `DEFAULT_IMPORTS` set in compiler-defined order so default imports are deterministic too.

### DIFF
--- a/compiler/src/main/java/play/japi/twirl/compiler/TwirlCompiler.java
+++ b/compiler/src/main/java/play/japi/twirl/compiler/TwirlCompiler.java
@@ -8,6 +8,8 @@ import java.io.File;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


mkurz:
@Mergifyio backport main 1.6.x

